PR #11, opened 2017-05-02
Expand macros to remove gotos in preparation for ARC conversion

miburf:
Preparation for converting project to ARC.
Fix methods whose goto action did more than return nil.
Expand macros.
Update macros to remove gotos.


mikenevi (approved):

namalu (approved):
This review looks good as the next step towards moving to ARC. It should be noted that the removal of the HVCLEAR Macro now creates many memory leaks where objects that were retained are no longer released. However, the next review that actually moves this code to ARC should resolve this issue.

PR #15, opened 2017-05-04
remove deprecated xml parser

mikenevi:
This takes out the deprecated xml parser from HVMobile.


MHVLib/HVMobile/HealthVaultRecord.m
```diff
@@ -76,30 +76,18 @@ - (id)initWithXml: (NSString *)xml
...
+    
+    self.recordName = [reader readString];
+    self.recordId = [reader readAttribute:@"id"];
+    self.authStatus = [reader readAttribute:@"app-record-auth-action"];
+    
+    if (!self.recordId) {
+        return NO;
+    } else {
+        return YES;
+    }
```
namalu: While we're modifying this file, lets make the code style consistent - move { to the next line in all cases. Also, it should be safe to remove the iVar definitions in the header file, the private interface definition, and the synthesize statements.

MHVLib/HVMobile/HealthVaultResponse.m
```diff
@@ -18,18 +18,12 @@
```
namalu, line 18: I recommend the same cleanup in this class to... Move { to the next line in all cases. Also, it should be safe to remove the iVar definitions in the header file, the private interface definition, and the synthesize statements.

MHVLib/HVMobile/HealthVaultResponse.m
```diff
@@ -18,18 +18,12 @@
...
 
 -(BOOL) deserializeXml:(NSString *) xml;
 
-/// Initializes the fields using xml string provided.
-/// @param xml - response xml representation.
-- (BOOL)parseFromXml: (NSString *)xml __deprecated; // Use deserializeXml instead
-
 /// Retrieves info section from xml.
 /// Info section is represented by <wc:info> xml element.
 /// @param xml - xml from which to retrieve info section.
```
namalu: There is some code that is on line 42 that is commented out that can now be deleted.

MHVLib/HVMobile/HealthVaultService.m
```diff
@@ -324,22 +323,11 @@ - (NSString *)getCastCallInfoSection {
 }
 
 - (void)saveCastCallResults: (NSString *)responseXml {
-
-    @autoreleasepool
-    {
-        @try
-        {
```
namalu: I think we might want to leave the try/catch here, but add a call to HVASSERT_MESSAGE in the catch so we can get a log if the deserialization fails and throws. At some point, we want to clean up all the try/catch code, but for now it seems safer to leave this one.

namalu (changes requested):

namalu (approved):

PR #19, opened 2017-05-05
Updating to modern Objective-C syntax

miburf:
Modernizing & cleanup code.
Remove synthesize
Removing m_ from instance variables
Add instancetype to inits


MHVLib/MHVUser.h
```diff
@@ -24,80 +24,71 @@
...
 
-//-------------------------
+// -------------------------
 //
 // Methods
 //
-//-------------------------
+// -------------------------
 
--(id) initFromLegacyRecords:(NSArray *) recordArray;  // Infrastructure - will eventually go away
```
namalu, line 70: Let's leave this comment in here until we can determine if this has been deprecated.

namalu (changes requested):

namalu (approved):

PR #20, opened 2017-05-08
Updated MHVCollecion to NSObject to prevent issues arising from acces…

namalu:
Updated MHVCollection to inherit from NSObject to prevent issues that can happen from accessing 'outer" array.


MHVLib/Collections/MHVCollection.h
```diff
@@ -18,16 +18,39 @@
...
 #import <Foundation/Foundation.h>
-
+#import "MHVBlock.h"
 
 //
 // Collections allow objects of a particular type only
 // They enforce the type
 //
-@interface MHVCollection : NSMutableArray
+@interface MHVCollection : NSObject
```
miburf: Already has the method in the .m file, so can add NSFastEnumeration

miburf (approved):

PR #24, opened 2017-05-08
Updating to modern Objective-C syntax

miburf:
Modernizing & cleanup code.
Remove synthesize
Removing m_ from instance variables
Add instancetype to inits


MHVLib/Types/MHVName.m
```diff
@@ -19,141 +19,139 @@
...
-    self.first = first;
-    self.middle = middle;
-    self.last = last;
-    
-    [self buildFullName];
-    MHVCHECK_NOTNULL(m_full);
-    
+    if (self)
+    {
+        _first = first;
```
mikenevi, line 44: Seems odd. Use the self.x instead?

MHVLib/Types/MHVNameValue.m
```diff
@@ -19,132 +19,126 @@
...
     self = [super init];
-    MHVCHECK_SELF;
-    
-    self.name = name;
-    self.value = value;
-    
+    if (self)
+    {
+        _name = name;
+        _value = value;
```
mikenevi, line 47: here too

mikenevi (approved):

namalu (approved):

MHVLib/Types/MHVName.m
miburf: It's better to use _x in init methods, to not trigger other setter code

PR #30, opened 2017-05-10
MHVHttpService, to replace deprecated NSURLConnection

miburf:
Add MHVHttpService, to replace NSURLConnection and use NSURLSession
Methods to prepare for RESTful endpoints


MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,271 @@
...
+    NSURLRequest *request = [self requestWithUrl:url
+                                            data:data
+                                     chunkOffset:chunkOffset
+                                       totalSize:blobSource.length];
+    
+    [Logger write:[NSString stringWithFormat:@"Blob upload chunk at offset %li", (long)chunkOffset]];
+
+//    [[self.urlSession uploadTaskWithRequest:request
+//                                   fromData:data
+//                          completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error)
```
mikenevi: remove

MHVLib/Types/Blob/MHVBlobSource.h
```diff
@@ -29,7 +29,7 @@
 
 @property (readonly, nonatomic) NSUInteger length;
 
-- (NSData *)readStartAt:(int)offset chunkSize:(int)chunkSize;
+-(NSData *) readStartAt:(NSInteger) offset chunkSize:(NSInteger) chunkSize;
```
mikenevi, line 32: spacing here isn't standard

mikenevi (approved):

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+                                               delegateQueue:nil];
+        
+        _certificateCheckQueue = [[NSOperationQueue alloc] init];
+    }
+    
+    return self;
+}
+
+- (void)sendRequestForURL:(NSURL *)url
+                 withData:(NSString *)dataString
```
namalu: Let's try to be a little more clear as to what 'data' is in the method signature. I'm assuming it's the body of the request? Maybe sendRequestForUrl:body:completion? Also, let's remove the 'with'.

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+    {
+        [request setValue:headers[key] forHTTPHeaderField:key];
+    }
+    
+    NSInteger currentRequest = (++self.requestCount);
+    NSDate *startDate = [NSDate date];
+
+    [Logger write:[NSString stringWithFormat:@"Begin request #%li", (long)currentRequest]];
+
+    //TODO: Could return NSURLSessionTask to allow tasks to be cancelled; that works for requests, but not blob uploads
```
namalu: We should leave the door open for cancellation. If this means that we are going to need to expose a separate method for uploading blobs, then we should do that. In other words, yes - Let's return the NSURLSessionTask.

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+    [[self.urlSession dataTaskWithRequest:(NSURLRequest *)request
+                        completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error)
+      {
+          NSInteger statusCode = ((NSHTTPURLResponse *)response).statusCode;
+          
+          [Logger write:[NSString stringWithFormat:@"Response for #%li has status code: %li (%0.4f seconds)",
+                         (long)currentRequest, (long)statusCode, [[NSDate date] timeIntervalSinceDate:startDate]]];
+          
+          if (error)
+          {
```
namalu, line 108: I see that completion is _Nonnull, but let's code defensively here - you should check if completion is nil to prevent this from crashing.

MHVLib/HTTP/MHVHttpServiceProtocol.h
```diff
@@ -0,0 +1,80 @@
...
+ */
+- (void)sendRequestForURL:(NSURL *)url
+                 withData:(NSString *)dataString
+               completion:(void (^)(MHVHttpServiceResponse *_Nullable response, NSError *_Nullable error))completion;
+
+/**
+ Send a request to HealthVault service
+
+ @param url the endpoint for the request
+ @param dataString data to send as POST body (usually XML)
```
namalu: Delete '(usually XML)'. We don't know how the consumers of the SDK will be using it.

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+@interface MHVHttpService () <NSURLSessionDelegate>
+
+@property (nonatomic, strong) NSURLSession *urlSession;
+@property (nonatomic, strong) NSOperationQueue *certificateCheckQueue;
+
+@property (nonatomic, assign) NSInteger requestCount;
+
+@end
+
+@implementation MHVHttpService
```
namalu, line 35: We need to code defensively. Though the compiler warns if a non-null value is being passed into a method, it does not prevent the code from being compiled. We need to check if parameters are nil and handle them appropriately.

MHVLib/HTTP/MHVHttpServiceProtocol.h
```diff
@@ -0,0 +1,80 @@
...
+                 withData:(NSString *)dataString
+                  headers:(NSDictionary<NSString *, NSString *> *_Nullable)headers
+               completion:(void (^)(MHVHttpServiceResponse *_Nullable response, NSError *_Nullable error))completion;
+
+/**
+ Download a file from HealthVault service to a local file path
+
+ @param url the endpoint for the request
+ @param path local file path where the downloaded file will be stored
+        For security, the file's protection attributes will be set to NSFileProtectionComplete
```
namalu: This setting cannot be used if we want to support operations in the background. We should consider using NSFileProtectionCompleteUntilFirstUserAuthentication.

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+
+- (instancetype)init
+{
+    self = [super init];
+    if (self)
+    {
+        _urlSession = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration]
```
namalu, line 42: This should be injected into this class so it can be tested.

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+{
+    NSInteger statusCode = ((NSHTTPURLResponse *)response).statusCode;
+    
+    return [[MHVHttpServiceResponse alloc] initWithResponseData:data
+                                               statusCode:statusCode];
+}
+
+#pragma mark - NSURLSessionDelegate
+
+- (void)     URLSession:(NSURLSession *)session
```
namalu: The extra space here just looks bad, I know we are trying to line up the colons, but this just doesn't look right.

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+    [self uploadBlobSource:blobSource toUrl:url chunkOffset:0 chunkSize:chunkSize completion:completion];
+}
+
+- (void)uploadBlobSource:(id<MHVBlobSource>)blobSource
+                   toUrl:(NSURL *)url
+             chunkOffset:(NSUInteger)chunkOffset
+               chunkSize:(NSUInteger)chunkSize
+              completion:(void (^)(MHVHttpServiceResponse *_Nullable response, NSError *_Nullable error))completion
+{
+    MHVASSERT_PARAMETER([url.scheme isEqualToString:@"https"]);
```
namalu: The MHVASSERT_PARAMETER() macro is used to assert that a parameter is not nil. You shouldn't use it to validate the state of an object. You can just use MHVASSERT()

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+{
+    [self sendRequestForURL:url withData:dataString headers:nil completion:completion];
+}
+
+- (void)sendRequestForURL:(NSURL *)url
+                 withData:(NSString *)dataString
+                  headers:(NSDictionary<NSString *, NSString *> *_Nullable)headers
+               completion:(void (^)(MHVHttpServiceResponse *_Nullable response, NSError *_Nullable error))completion
+{
+    MHVASSERT_PARAMETER([url.scheme isEqualToString:@"https"]);
```
namalu: The MHVASSERT_PARAMETER() macro is used to assert that a parameter is not nil. You shouldn't use it to validate the state of an object. You can just use MHVASSERT()

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,276 @@
...
+                                                              NSFileProtectionKey : NSFileProtectionComplete
+                                                              }
+                                               ofItemAtPath:path
+                                                      error:nil];
+              completion(nil);
+          }
+      }] resume];
+}
+
+- (void)uploadBlobSource:(id<MHVBlobSource>)blobSource
```
namalu: We should update the name of the protocol to MHVBlobSourceProtocol

MHVLib/HTTP/MHVHttpServiceResponse.h
```diff
@@ -0,0 +1,44 @@
...
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MHVHttpServiceResponse : NSObject
+
+// The response data
+@property (nonatomic, strong, readonly, nullable) NSData *responseAsData;
+@property (nonatomic, strong, readonly, nullable) NSString *responseAsString;
+@property (nonatomic, strong, readonly, nullable) NSDictionary *responseAsDictionary;
```
namalu: Do we need the responseAsDictionary property? It would seem that all responses (other than blobs) will be deserialized into an object.

MHVLib/HVMobile/HealthVaultRequest.m
```diff
@@ -22,99 +22,62 @@
...
 
--(NSURLConnection *)connection
+- (instancetype)initWithMethodName:(NSString *)name
+                     methodVersion:(float)methodVersion
+                       infoSection:(NSString *)info
+                            target:(NSObject *)target
+                          callBack:(SEL)callBack
 {
-    @synchronized(self)
+    if (self = [super init])
```
namalu: Let's format this like this:

self = [super init];

if (self)...

namalu (changes requested):

MHVLib/HTTP/MHVHttpService.h
```diff
@@ -1,8 +1,8 @@
 //
-//  MHVHttpTransport.m
-//  MHVLib
+// MHVHttpService.h
+// MHVLib
 //
-// Copyright 2017 Microsoft Corp.
+//  Copyright (c) 2017 Microsoft Corporation. All rights reserved.
```
anbutle: test comment

MHVLib/HTTP/MHVHttpService.m
```diff
@@ -0,0 +1,275 @@
...
+          }
+          else
+          {
+              [[NSFileManager defaultManager] setAttributes:@{
+                                                              NSFileProtectionKey : NSFileProtectionComplete
+                                                              }
+                                               ofItemAtPath:path
+                                                      error:nil];
+              completion(nil);
+          }
```
anbutle: another test comment

anbutle (commented):
testing comment aggregation

MHVLib/HTTP/MHVHttpService.m
miburf: added many more checks

MHVLib/HTTP/MHVHttpService.m
miburf: Added initializer so NSURLSession can be passed in, and init that creates a default (NSURLSession has a delegate that can only be set when it's created)

MHVLib/HTTP/MHVHttpServiceResponse.h
miburf: The JSON parser will need a dictionary as we merge the deserializer from the Insights app, and the XML parser currently uses the string.

MHVLib/HTTP/MHVHttpService.m
miburf: added checks for completion

namalu (approved):

PR #33, opened 2017-05-11
Update blob storage to use MHVHttpService

miburf:
Remove NSUrlConnection based requests for Blob storage
Add MHVHttpTasks for observing progress and cancelling requests
Update sample app
Add tests for MHVHttpService


MHVLib/MHVAsyncTask.h
```diff
@@ -62,7 +62,7 @@ typedef void (^MHVTaskCompletion) (MHVTask *task);
 //
 // This task's parent task
 //
-@property (readonly, nonatomic, weak) MHVTask *parent;
+@property (readonly, nonatomic, strong) MHVTask *parent;
```
miburf, line 65: parent isn't run until child completes; ARC changes means the parent got deallocated before the child told it to start.

namalu (approved):

PR #35, opened 2017-05-12
Check for blob data before serializing to XML

miburf:
Getter created blob object, check for blob data before adding to XML


MHVLib/Types/MHVItem.m
```diff
@@ -90,7 +90,7 @@ - (void)setNote:(NSString *)note
 
 - (MHVBlobPayload *)blobs
 {
-    MHVENSURE(_blobs, [MHVBlobPayload class]);
+    MHVENSURE(_blobs, MHVBlobPayload);
```
namalu, line 93: We should get rid of this macro - It hides some pretty important code. We don't need to remove it everywhere in the project right now, but we should start deleting it when we touch a file.

namalu (approved):

PR #45, opened 2017-05-18
Fixes for token refresh

miburf:
Add forceRefresh to refreshSessionCredential to refresh token.
Don't add authSession header if method is anonymous

### DIFF
--- a/MHVLib/Connection/MHVConnection.m
+++ b/MHVLib/Connection/MHVConnection.m
@@ -33,6 +33,7 @@
 #import "MHVSessionCredentialClientProtocol.h"
 #import "MHVPlatformClient.h"
 #import "MHVPersonClient.h"
+#import "MHVValidator.h"
 
 static NSString *const kCorrelationIdContextKey = @"WC_CorrelationId";
 static NSString *const kResponseIdContextKey = @"WC_ResponseId";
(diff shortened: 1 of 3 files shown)


MHVLib/Connection/MHVConnection.m
```diff
@@ -134,13 +134,12 @@ - (void)executeMethod:(MHVMethod *_Nonnull)method
...
 - (void)getPersonInfoWithCompletion:(void (^_Nonnull)(MHVPersonInfo *_Nullable, NSError *_Nullable error))completion;
 {
-    
 }
 
 - (void)authenticateWithViewController:(UIViewController *_Nullable)viewController
                             completion:(void(^_Nullable)(NSError *_Nullable error))completion;
 {
-    
+    NSAssert(NO, @"Subclasses must implement %@", NSStringFromSelector(_cmd));
```
namalu: Let's use the MHV macros here - They log, rather then crash.

MHVLib/Connection/MHVSodaConnection.m
```diff
@@ -378,9 +378,12 @@ - (void)setServiceInstanceWithInstanceId:(NSString *)instanceId
...
 }
 
-- (void)refreshSessionCredentialWithCompletion:(void(^_Nullable)(NSError *_Nullable error))completion
+- (void)refreshSessionCredential:(BOOL)forceRefresh completion:(void(^_Nullable)(NSError *_Nullable error))completion
 {
-    if (self.sessionCredential)
+    self.credentialClient.connection = self;
+    self.credentialClient.sharedSecret = self.applicationCreationInfo.sharedSecret;
+    
+    if (self.sessionCredential && !forceRefresh)
```
namalu: I think we should just remove this check altogether. If the session credential is stale (after retrieving it from the keychain) we will need to refresh it anyways - We should just always refresh the credential if "authenticate" is called.

namalu (approved):

PR #46, opened 2017-05-18
Implement ThingClient

miburf:
Implement ThingClient
Update SDKFeatures to use new thing client
Add multi-record configuration and record selector to SDKFeatures sample


MHVLib/Clients/MHVThingClientProtocol.h
```diff
@@ -18,47 +18,62 @@
...
 #import "MHVClientProtocol.h"
 
-@class MHVThing, MHVThingQuery, MHVThingCollection;
+@class MHVThing, MHVThingQuery, MHVThingCollection, MHVThingQueryCollection, MHVThingQueryResultCollection;
 
 NS_ASSUME_NONNULL_BEGIN
 
 @protocol MHVThingClientProtocol <NSObject>
 
+/**
```
namalu, line 27: This is a 'public' protocol that consumers will interface with - You are going to have add documentation for all of the methods.

MHVLib/Connection/MHVSodaConnection.m
```diff
@@ -415,6 +417,33 @@ - (void)refreshSessionCredentialWithCompletion:(void(^_Nullable)(NSError *_Nulla
     }];
 }
 
+- (void)getPersonInfoWithCompletion:(void (^)(MHVPersonInfo * _Nullable, NSError * _Nullable))completion
```
namalu: 3 things:

1. This method has been stubbed in the super class (MHVConnection), be sure to add an assert there to make sure this method is implemented.

2. This is misplaced under the 'Private' pragma mark.

MHVLib/Connection/MHVSodaConnection.m
```diff
@@ -500,6 +535,12 @@ - (void)setConnectionPropertiesFromKeychain
     if (!self.personInfo)
     {
         self.personInfo = [self.keychainService xmlObjectForKey:kPersonInfoKey];
+
+        if (!self.personInfo.selectedRecordID ||
+            [self.personInfo.selectedRecordID isEqual:[[NSUUID alloc] initWithUUIDString:kBlankUUID]])
+        {
+            self.personInfo.selectedRecordID = [self.personInfo.records firstObject].ID;
+        }
```
namalu: Rather than do this in 2 places, I think we should just set the personInfo.records property before saving to the keychain.

MHVLib/Connection/MHVSodaConnection.m
```diff
@@ -513,6 +554,11 @@ - (void)clearConnectionProperties
     _personInfo = nil;
 }
 
+- (void)clearSessionCredential
+{
+    _sessionCredential = nil;
+}
```
namalu: It doesn't appear that anyone is calling this.

MHVLib/Errors/MHVErrorConstants.h
```diff
@@ -28,6 +28,7 @@ typedef NS_ENUM(NSUInteger, MHVErrorType)
     MHVErrorTypeIOError,
     MHVErrorTypeUnauthorized,
     MHVErrorTypeOperationCancelled,
+    MHVErrorTypeNoResults,
```
namalu: I don't think no results is an error. If a request yields no results, we should just return the empty collection. If we return an error for no results, developers will need create a special execution path in their error flow to handle the case where there is no error, but rather no results. We should reserve errors for actual problems.

namalu (approved):

PR #47, opened 2017-05-18
Personal/namalu/add client factory

namalu:
I added a client factory to improve the testability of the connection classes. I also added code to clear caches and cookies from the UIWebView so the auth session would not be retained.

### DIFF
--- a/MHVLib/Clients/MHVClientFactory.h
+++ b/MHVLib/Clients/MHVClientFactory.h
@@ -1,0 +1,25 @@
+//
+//  MHVClientFactory.h
+//  MHVLib
+//
+//  Created by Nathan Malubay on 5/18/17.
+//  Copyright © 2017 Microsoft Corporation. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol MHVConnectionProtocol, MHVPersonClientProtocol, MHVPlatformClientProtocol, MHVThingClientProtocol, MHVVocabularyClientProtocol, MHVSessionCredentialClientProtocol;
+
+@interface MHVClientFactory : NSObject
+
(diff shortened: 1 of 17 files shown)


MHVLibTests/Connection/MHVSodaConnectionTests.m
```diff
@@ -0,0 +1,102 @@
+//
+//  MHVSodaConnectionTests.m
+//  MHVLib
+//
+//  Created by Nathan Malubay on 5/18/17.
+//  Copyright © 2017 Microsoft Corporation. All rights reserved.
```
miburf, line 6: License

MHVLib/Clients/MHVClientFactory.h
```diff
@@ -0,0 +1,25 @@
+//
+//  MHVClientFactory.h
+//  MHVLib
+//
+//  Created by Nathan Malubay on 5/18/17.
```
miburf: License

MHVLib/Clients/MHVClientFactory.h
```diff
@@ -0,0 +1,25 @@
...
+//  Copyright © 2017 Microsoft Corporation. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol MHVConnectionProtocol, MHVPersonClientProtocol, MHVPlatformClientProtocol, MHVThingClientProtocol, MHVVocabularyClientProtocol, MHVSessionCredentialClientProtocol;
+
```
miburf, line 22: NS_ASSUME_NONNULL_BEGIN, since nothing should be nil

miburf (approved):

PR #49, opened 2017-05-19
Standardize completion blocks

mikenevi:


namalu (approved):
We spoke offline about formatting curly brackets, if it's not a big deal let's do that and merge this back into develop.

namalu (approved):

PR #51, opened 2017-05-19
Implement retrieving blob data

miburf:
Implement retrieving blob data
Added to MHVMethod and MHVConnection to do HTTP methods


MHVLib/Methods/MHVMethod.h
```diff
@@ -38,6 +46,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign) int version;
 
 /**
+ A MHVMethodType indicating the method type
+ */
+@property (nonatomic, assign, readonly) MHVMethodType type;
+
+/**
```
miburf, line 53: I debated whether to do this as a type or subclasses of MHVMethod: MHVXMLMethod, MHVRestMethod, MHVDataMethod, etc.  I may still do that...

mikenevi (approved):

MHVLib/Clients/MHVPersonClient.m
```diff
@@ -120,4 +123,60 @@ - (void)getAuthorizedRecordsWithRecordIds:(NSArray<NSUUID *> *)recordIds
...
+    }
+    
+    //Get the personalImage thing, including the blob section
+    MHVThingQuery *query = [[MHVThingQuery alloc] initWithTypeID:MHVPersonalImage.typeID];
+    query.view.sections = MHVThingSection_Blobs;
+    
+    [self.connection.thingClient getThingsWithQuery:query
+                                           recordId:recordId
+                                         completion:^(MHVThingCollection * _Nullable things, NSError * _Nullable error)
+     {
```
namalu, line 141: We should check if the completion is nil and return before doing anything else.

MHVLib/Clients/MHVPersonClient.m
```diff
@@ -120,4 +123,60 @@ - (void)getAuthorizedRecordsWithRecordIds:(NSArray<NSUUID *> *)recordIds
...
+         //Gets the defaultBlob from the first thing in the result collection
+         MHVThing *thing = [things firstObject];
+         if (!thing)
+         {
+             completion(nil, [NSError MHVNotFound]);
```
namalu, line 146: According to the HealthVault developer documentation, the call returns a collection of images that will contain either a single image or no images (https://msdn.microsoft.com/en-us/healthvault/dn800726). I think returning an error here deviates from the pattern. In this case I think it's appropriate to return nil for both the error and the image just like the documentation in the protocol states.

MHVLib/Clients/MHVThingClient.m
```diff
@@ -418,6 +418,175 @@ - (void)removeThings:(MHVThingCollection *)things
...
+    if (!completion)
+    {
+        return;
+    }
+    
+    if (!thing || !recordId)
+    {
+        if (completion)
+        {
+            completion(nil, [NSError MVHInvalidParameter]);
```
namalu, line 445: Let's add a detailed description to errors, so it is clear to developers what is going on. It will also help if people are logging errors.

MHVLib/Methods/MHVMethod.h
namalu: I'm don't think these are methods at all, and should not be subclassed or added to the MHVMethod class. There is a list of methods on the developer site (https://developer.healthvault.com/Methods), we should avoid adding things to this class that are not methods listed in the dev portal.

namalu (changes requested):
We can talk more about the blob stuff in our 2:30 meeting today - Let's get feedback from Surya and David before making any changes.

miburf:
Pulling for now to align with OneSDK...

PR #52, opened 2017-05-22
Added More MHVConnection class tests. Added authorize additional reco…

namalu:
…rds in SDKFeatures app.

### DIFF
--- a/Samples/SDKFeatures/SDKFeatures/MHVFeatureActions.h
+++ b/Samples/SDKFeatures/SDKFeatures/MHVFeatureActions.h
@@ -21,7 +21,7 @@
 
 @interface MHVFeatureActions : NSObject
 
-- (id)initWithTitle:(NSString *)title;
+- (instancetype)initWithTitle:(NSString *)title;
 
 - (BOOL)addFeature:(NSString *)title andAction:(MHVAction)action;
 
--- a/Samples/SDKFeatures/SDKFeatures/MHVFeatureActions.h
+++ b/Samples/SDKFeatures/SDKFeatures/MHVFeatureActions.h
@@ -19,17 +19,12 @@
 
 #import <UIKit/UIKit.h>
 
-@interface MHVFeatureActions : NSObject<UIActionSheetDelegate>
-{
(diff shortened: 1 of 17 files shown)


Samples/SDKFeatures/SDKFeatures/MHVFeatureActions.h
```diff
@@ -19,17 +19,12 @@
...
-@interface MHVFeatureActions : NSObject<UIActionSheetDelegate>
-{
-@private
-    UIActionSheet* m_actionSheet;
-    NSMutableArray* m_actions;
-}
+@interface MHVFeatureActions : NSObject
 
--(id) initWithTitle:(NSString *) title;
+- (id)initWithTitle:(NSString *)title;
```
miburf: nit, instancetype. In the .m file too

Samples/SDKFeatures/SDKFeatures/MHVMoreFeatures.m
```diff
@@ -90,9 +90,62 @@ -(void)getServiceDefinition
...
+             @"[ShellRedirect]", serviceDefinition.shell.redirectUrl,
+             @"[Example Config Entries]",
+             configEntry.key, @"==", configEntry.value, @"==========",
+             configEntry2.key, @"==", configEntry2.value];
+            
+            [MHVUIAlert showInformationalMessage:output];
+        }
+        
+        
+        [self.listController.statusLabel clearStatus];
```
miburf, line 124: need to dispatch to main thread.  Should we have a setting on the SDK to do the completions on the main thread?  Could affect performance when set, but for a simple app it would also ease development using the SDK.

miburf (approved):

Samples/SDKFeatures/SDKFeatures/MHVMoreFeatures.m
namalu: I don't think we should do that just yet. Most developers should understand that completion blocks can return on a different thread, and they should code appropriately.

PR #55, opened 2017-05-23
Update executeMethod to support both Method and Rest operations

miburf:
Update executeMethod to support both Method and Rest operations.
MHVHttpServiceRequest can hold either type, can refresh token and retry using common logic in MHVConnection

### DIFF
--- a/MHVLib/Clients/MHVClientFactory.m
+++ b/MHVLib/Clients/MHVClientFactory.m
@@ -20,6 +20,7 @@
 #import "MHVPersonClient.h"
 #import "MHVPlatformClient.h"
 #import "MHVSessionCredentialClient.h"
+#import "MHVThingClient.h"
 
 @implementation MHVClientFactory
 
@@ -35,11 +36,12 @@
 
 - (id<MHVThingClientProtocol>)thingClientWithConnection:(id<MHVConnectionProtocol>)connection
 {
-    return nil;
+    return [[MHVThingClient alloc] initWithConnection:connection];
 }
 
 - (id<MHVVocabularyClientProtocol>)vocabularyClientWithConnection:(id<MHVConnectionProtocol>)connection
 {
+    NSAssert(NO, @"Must Implement");
     return nil;
 }
 
--- a/MHVLib/Clients/MHVClientFactory.m
+++ b/MHVLib/Clients/MHVClientFactory.m
@@ -41,7 +41,6 @@
 
 - (id<MHVVocabularyClientProtocol>)vocabularyClientWithConnection:(id<MHVConnectionProtocol>)connection
 {
-    NSAssert(NO, @"Must Implement");
     return nil;
 }
 
--- a/MHVLib/Clients/MHVRemoteMonitoringClient.m
+++ b/MHVLib/Clients/MHVRemoteMonitoringClient.m
@@ -57,7 +57,7 @@
                            formParams:(NSDictionary<NSString *, NSString *> *_Nullable)formParams
                                  body:(NSData *_Nullable)body
                               toClass:(Class)toClass
-                           completion:(void(^_Nonnull)(id _Nullable output, NSError *_Nullable error))completion
+                           completion:(void(^_Nullable)(id _Nullable output, NSError *_Nullable error))completion
 {
     //    [[MHVRemoteMonitoringClient current] requestWithPath:path
     //                                              method:method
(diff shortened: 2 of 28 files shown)


MHVLib/Clients/MHVClientFactory.m
```diff
@@ -35,11 +36,12 @@ @implementation MHVClientFactory
 
 - (id<MHVThingClientProtocol>)thingClientWithConnection:(id<MHVConnectionProtocol>)connection
 {
-    return nil;
+    return [[MHVThingClient alloc] initWithConnection:connection];
 }
 
 - (id<MHVVocabularyClientProtocol>)vocabularyClientWithConnection:(id<MHVConnectionProtocol>)connection
 {
+    NSAssert(NO, @"Must Implement");
```
namalu: This is going to cause a merge conflict with Andrew. Let's not do this.

MHVLib/Clients/MHVRemoteMonitoringClient.m
```diff
@@ -18,61 +18,109 @@
...
+- (void)requestWithPath:(NSString *_Nonnull)path
+                 method:(NSString *_Nonnull)method
+             pathParams:(NSDictionary<NSString *, NSString *> *_Nullable)pathParams
+            queryParams:(NSDictionary<NSString *, NSString *> *_Nullable)queryParams
+             formParams:(NSDictionary<NSString *, NSString *> *_Nullable)formParams
+                   body:(NSData *_Nullable)body
+                toClass:(Class)toClass
+             completion:(void(^_Nonnull)(id _Nullable output, NSError *_Nullable error))completion
+{
+    if (!completion)
```
namalu: There may be some cases where consumers of the SDK don't require a completion, we should make the completion optional.

MHVLib/Connection/MHVConnection.m
```diff
@@ -90,7 +87,7 @@ - (NSUUID *_Nullable)applicationId;
     return nil;
 }
 
-- (void)executeMethod:(MHVMethod *_Nonnull)method
+- (void)executeMethod:(id<MHVHttpServiceOperationProtocol> _Nonnull)method
```
namalu: I think we should update to executeHttpServiceOperation:

MHVLib/Clients/MHVRemoteMonitoringClient.m
```diff
@@ -18,61 +18,109 @@
...
+    {
+        _connection = connection;
+    }
+    
+    return self;
+}
+
+//PENDING: Swagger generated code needs to update to be injected with RemoteMonitoringClient & remove this method
++ (NSURLSessionTask *)requestWithPath:(NSString *_Nonnull)path
+                               method:(NSString *_Nonnull)method
```
namalu, line 54: we should avoid using the term 'method' here. It's important to differentiate the difference between the legacy methods and the new rest 'apis'. The develop documentation on the HealthVault website refers to t these as 'REST API'. This might be something Mike has to do in the swagger gen templates.

MHVLib/Methods/MHVRestRequest.h
```diff
@@ -0,0 +1,53 @@
...
+    //MHVRestRequestDownloadBlobFile,
+    //MHVRestRequestUploadBlob,
+};
+
+@interface MHVRestRequest : NSObject <MHVHttpServiceOperationProtocol>
+
+@property (nonatomic, assign, readonly)           MHVRestRequestType restRequestType;
+
+@property (nonatomic, strong, readonly)           NSString      *path;
+@property (nonatomic, strong, readonly)           NSString      *method;
```
namalu: let's not use method here - api would be more appropriate.

MHVLib/Connection/MHVConnection.m
```diff
@@ -191,14 +207,68 @@ - (void)executeMethodRequest:(MHVMethodRequest *)request
...
+
+- (void)executeRestRequest:(MHVHttpServiceRequest *)request
+{
+    MHVRestRequest *restRequest = request.serviceOperation;
     
-    
+    if (restRequest.restRequestType == MHVRestRequestJSON)
+    {
+        [self executeJsonRestRequest:request];
+    }
```
namalu: I'm not sure we should be doing this just yet - We should probably wait until after we determine the best approach for blobs before adding code to handle it.

namalu (approved):
Just a couple of changes requested.

MHVLib/Methods/MHVRestRequest.h
miburf: api seemed odd there too.  Changed to httpMethod which matches what it is called in NSURL

MHVLib/Clients/MHVRemoteMonitoringClient.m
miburf: This one Mike will have to change in the swagger templates, then can update here

PR #56, opened 2017-05-24
Update API templates to use MHVRestRequest methods.

mikenevi:


MHVLib/Models/Generated/Categories/MHVGoalsApi.m
```diff
@@ -62,45 +62,37 @@ + (NSURLSessionTask*)createGoalsWithGoalsWrapper:(MHVGoalsWrapper*)goalsWrapper
...
-                                    }
-                                }
-            ];
+    
+    [self requestWithPath:resourcePath
+               httpMethod:@"POST"
+               pathParams:pathParams
+              queryParams:queryParams
+               formParams:formParams
+                     body:bodyParam
```
miburf, line 82: body is now NSData, which is what the iOS http service uses, so might need to add conversions here

MHVLib/Models/Generated/Categories/MHVGoalsApi.m
```diff
@@ -326,45 +288,37 @@ + (NSURLSessionTask*)patchGoalsWithGoalsWrapper:(MHVGoalsWrapper*)goalsWrapper
...
-                                        completion((MHVGoalsResponse*)data, error);
-                                    }
-                                }
-            ];
+    
+    [self requestWithPath:resourcePath
+               httpMethod:@"PATCH"
+               pathParams:pathParams
+              queryParams:queryParams
+               formParams:formParams
```
miburf, line 307: Nothing uses the formParams now.  If you don't think it's going to be used, can delete

miburf (approved):

MHVLib/Models/Generated/Categories/MHVGoalsApi.m
mikenevi: I'm not sure if it will be used or not. Will leave for now and remove later.

MHVLib/Models/Generated/Categories/MHVGoalsApi.m
mikenevi: I see. Will need to update this when I get to testing functions that use the body. That will come later, not as part of this review.

PR #58, opened 2017-05-31
Add tests for MHVConnection

miburf:
Add tests for MHVConnection for both HealthVault Method and Rest requests

### DIFF
--- a/MHVLibTests/Connection/MHVConnectionTests.m
+++ b/MHVLibTests/Connection/MHVConnectionTests.m
@@ -1,0 +1,264 @@
+//
+//  MHVConnectionTests.m
+//  MHVLib
+//
+// Copyright (c) 2017 Microsoft Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


mikenevi (approved):

MHVLibTests/Connection/MHVConnectionTests.m
```diff
@@ -0,0 +1,264 @@
...
+    testConnection.serviceInstance = [[MHVInstance alloc] init];
+    testConnection.serviceInstance.healthServiceUrl = [NSURL URLWithString:@"https://test.url"];
+    
+    testConnection.sessionCredential = [[MHVSessionCredential alloc] initWithToken:kDefaultToken sharedSecret:kDefaultSharedSecret];
+    
+    // Requested values
+    __block NSURL *requestedURL;
+    __block NSString *httpMethod;
+    __block NSDictionary *requestedHeaders;
+    __block NSData *requestedBody;
```
namalu: I think you are going to need to set these values to a default state beforeEach() - Otherwise you may end up with false positives/negatives from a leftover state.

MHVLibTests/Connection/MHVConnectionTests.m
```diff
@@ -0,0 +1,264 @@
...
+                beforeAll(^{
+                    MHVMethod *method = [MHVMethod getThings];
+                    method.parameters = @"GETTHINGSBODY";
+                    [testConnection executeHttpServiceOperation:method
+                                                     completion:^(MHVServiceResponse * _Nullable response, NSError * _Nullable error) { }];
+                });
+                
+                it(@"sendRequest should have been performed", ^
+                   {
+                       [[requestedURL shouldEventually] beKindOfClass:[NSURL class]];
```
namalu: I believe all of these need to be wrapped in expectFutureValue().

namalu (changes requested):

namalu (approved):

PR #59, opened 2017-05-31
Implementing VocabularyClient

anbutle:
1. Implemented vocabulary factory
2. Added vocabulary client unit tests
3. Refactored vocabulary objects to better align with cloud models
4. Updated vocabulary client protocol


MHVLib/Clients/MHVVocabularyClient.h
```diff
@@ -0,0 +1,37 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MHVVocabularyClient : NSObject<MHVVocabularyClientProtocol>
+
+- (instancetype)initWithConnection:(id<MHVConnectionProtocol>)connection;
+
+- (MHVMethod *) getVocabularySearchMethodWithSearchValue:(NSString *)searchValue
+                                           andSearchMode:(MHVSearchMode)searchMode
+                                           andMaxResults:(NSNumber * _Nullable)maxResults
+                                        andVocabularyKey:(MHVVocabularyKey * _Nullable)vocabularyKey;
```
namalu: You can delete this method definition from the header file. For methods that are called 'internally' you don't need to declare them in the header - This is the objc way of making them 'private'.

MHVLib/Clients/MHVVocabularyClient.m
```diff
@@ -0,0 +1,249 @@
...
+    [self.connection executeHttpServiceOperation:method completion:^(MHVServiceResponse *_Nullable response, NSError  *_Nullable error)
+    {
+        if (error)
+        {
+            completion(nil, error);
+            return;
+        }
+        
+        MHVVocabularyKeyCollection *vocabularyKeys = (MHVVocabularyKeyCollection*)[XSerializer newFromString:response.infoXml withRoot:@"info" andElementName:@"vocabulary-key" asClass:[MHVVocabularyKey class] andArrayClass:[MHVVocabularyKeyCollection class]];
+
```
namalu, line 69: We should do a check here to make sure that vocabularyKeys was parsed properly. There are a few examples of this in the MHVPlatformClient.m, where we check if the value is nil and provide an error if it is.

MHVLib/Clients/MHVVocabularyClient.m
```diff
@@ -0,0 +1,249 @@
...
+              cultureIsFixed:(BOOL)cultureIsFixed
+                  completion:(void(^)(MHVVocabularyCodeSet *_Nullable vocabulary, NSError *_Nullable error))completion
+{
+    MHVASSERT_PARAMETER(key);
+    MHVASSERT_PARAMETER(completion);
+    
+    if (!completion)
+    {
+        return;
+    }
```
namalu, line 99: Because 'key' is a required parameter, we should check if the parameter is nil and complete with an MVHInvalidParameter error.

MHVLib/Clients/MHVVocabularyClient.m
```diff
@@ -0,0 +1,249 @@
...
+}
+
+- (void)getVocabulariesWithVocabularyKeys:(MHVVocabularyKeyCollection *)vocabularyKeys
+                           cultureIsFixed:(BOOL)cultureIsFixed
+                               completion:(void(^)(MHVVocabularyCodeSetCollection* _Nullable vocabularies, NSError *_Nullable error))completion
+{
+    MHVASSERT_PARAMETER(vocabularyKeys);
+    MHVASSERT_PARAMETER(completion);
+    
+    if (!completion || !vocabularyKeys)
```
namalu: The only time we return without calling the completion is if the completion parameter is nil. The thinking behind this is that if no completion block is provided, there is no way to provide results to the caller - So we don't bother issuing the request. However, if the caller fails to provide vocabularyKeys, we need to call the completion with an error telling the caller that vocabularyKeys is a required parameter. We probably also need to make sure the collection is no empty. We never want a completion block to be 'orphaned'. In this method, if a caller fails to provide the vocabularyKeys parameter we will just return, and they will be waiting indefinitely for the completion to be called.

MHVLib/Clients/MHVVocabularyClient.m
```diff
@@ -0,0 +1,249 @@
...
+    
+    [self getVocabulariesWithVocabularyKeys:[[MHVVocabularyKeyCollection alloc]initWithArray:@[key]] cultureIsFixed:cultureIsFixed completion:^(MHVVocabularyCodeSetCollection * _Nullable vocabularies, NSError * _Nullable error)
+    {
+        if (error)
+        {
+            completion(nil, error);
+            return;
+        }
+        
+        if (!vocabularies || [vocabularies count] <= 0)
```
namalu: [vocabularies count] is an unsigned integer and can never be < zero.

MHVLib/Clients/MHVVocabularyClient.m
```diff
@@ -0,0 +1,249 @@
...
+        {
+            completion(nil, nil);
+            return;
+        }
+        
+        completion([vocabularies objectAtIndex:0], nil);
```
namalu: You can simplify this logic quite a bit. You can just do completion([vocabularies firstObject], nil);. If vocabularies is nil, the result will be nil; if vocabularies is empty, the result will be nil; if vocabularies has multiple objects, the result will be [vocabularies objectAtIndex:0].

MHVLib/Clients/MHVVocabularyClient.m
```diff
@@ -0,0 +1,249 @@
...
+    [self.connection executeHttpServiceOperation:method completion:^(MHVServiceResponse * _Nullable response, NSError * _Nullable error)
+     {
+         if (error)
+         {
+             completion(nil, error);
+             return;
+         }
+         
+         MHVVocabularyKeyCollection *vocabularyKeys = (MHVVocabularyKeyCollection*)[XSerializer newFromString:response.infoXml withRoot:@"info" andElementName:@"vocabulary-key" asClass:[MHVVocabularyKey class] andArrayClass:[MHVVocabularyKeyCollection class]];
+
```
namalu, line 208: Is the result type of the 'SearchVocabulary' Method MHVVocabularySearchResults type? I'm wondering if you actually need to add a new method to the serializer, or if you can just use the results type.

MHVLib/Clients/MHVVocabularyClient.m
```diff
@@ -0,0 +1,249 @@
...
+    [self.connection executeHttpServiceOperation:method completion:^(MHVServiceResponse * _Nullable response, NSError * _Nullable error)
+    {
+        if (error)
+        {
+            completion(nil, error);
+            return;
+        }
+        
+        MHVVocabularyCodeSetCollection *vocabularies = (MHVVocabularyCodeSetCollection*)[XSerializer newFromString:response.infoXml withRoot:@"info" andElementName:@"vocabulary" asClass:[MHVVocabularyCodeSet class] andArrayClass:[MHVVocabularyCodeSetCollection class]];
+
```
namalu, line 166: I think the return type for 'GetVocabulary' is MHVVocabularyGetResults type.

namalu (changes requested):

MHVLib/Clients/MHVVocabularyClient.m
anbutle: SearchVocabulary actually has two possible return types currently. Depending on the parameters provided to the call, two different things happen. If you provide a VocabularyKey, then the CodeItems for that key are searched and the result is a VocabularyCodeSet. If you do not provide a VocabularyKey then the call searches all VocabularyKeys and the result is VocabularyKeyInfo. I have reached out to the coreplat team because this is poor API design and we should have two distinct methods instead of a single method with variable functionality. see https://developer.healthvault.com/Methods/ResponseSchema?Name=SearchVocabulary&Version=1

namalu (approved):

PR #60, opened 2017-06-01
REST API errors, string array deserialization, Plan Task API implementation

mikenevi:


MHVLib/Connection/MHVConnection.m
```diff
@@ -257,7 +257,21 @@ - (void)executeRestRequest:(MHVHttpServiceRequest *)request
             return;
         }
         
-        if (error)
+        if (response.hasError)
```
mikenevi, line 260: if (error) is fine if there is a protocol error - timed out, network not available, etc. If there's a bad response - 400, 500, etc - then hasError is true and we need to look at more details. Two questions I have here - is it safe to put the whole responseAsString in the error? and maybe we should create something more meaningful than MHVUnknown error?

MHVLib/Models/MHVModelBase.m
```diff
@@ -212,6 +212,11 @@ + (void)setProperties:(NSObject<MHVDataModelProtocol> *)object fromDictionary:(N
                             
                             @synchronized (propertyParametersCache)
                             {
+                                if (!objectParameters)
+                                {
+                                    objectParameters = [NSString class];
```
mikenevi, line 217: This may have deeper implications. The problem now is that when compiled, objective C treats NSArray<NSString*> as just NSArray. It drops the type, so we don't have the object parameter type and string arrays end up as nil after deserialization. I'm not sure the deeper implications here - what happens of arrays of arrays of strings? Is this still ok for NSArray<NSNumber> ? I don't have good examples of those so I figured we'd cross that bridge when we come to it.

mikenevi: hmmm... github filtered my comment, so it doesn't make a lot of sense. I was trying to say I wasn't sure if this would work if the datatype were an NSArray of NSNumber. Same conclusion - I figured we'd revisit that if and when it's needed.

MHVLib/Connection/MHVConnection.m
namalu: 1. It should be fine to add the responseAsString to the error. 2. It probably would be a good idea to create a 'MHVNetworkError' or some other error type that is more specific to making a request to the HealthVault service.

Samples/SDKFeatures/SDKFeatures/ViewControllers/ActionPlanTasks/MHVActionPlanTaskDetailViewController.m
```diff
@@ -0,0 +1,136 @@
...
+- (IBAction)deleteTask:(id)sender;
+
+@end
+
+@implementation MHVActionPlanTaskDetailViewController
+
+- (instancetype)initWithTaskId:(NSString *)taskId
+{
+    self = [super init];
+    self.taskId = taskId;
```
namalu: Just a style preference...

self = [super init];

if (self)
{
    _taskId = taskId;
    _connection = ...
}

return self;

Samples/SDKFeatures/SDKFeatures/ViewControllers/ActionPlanTasks/MHVActionPlanTaskDetailViewController.m
```diff
@@ -0,0 +1,136 @@
...
+    
+    return self;
+}
+
+- (void)viewDidLoad
+{
+    [super viewDidLoad];
+    self.task = [[MHVActionPlanTaskInstance alloc] init];
+    
+    [self.navigationController.navigationBar setTranslucent:FALSE];
```
namalu: nit: 'NO' instead of 'FALSE'

namalu (approved):

MHVLib/Connection/MHVConnection.m
mikenevi: Sounds good to me

Samples/SDKFeatures/SDKFeatures/ViewControllers/ActionPlanTasks/MHVActionPlanTaskDetailViewController.m
mikenevi: right. done.

PR #61, opened 2017-06-01
Update Thing Client to fetch pending things

miburf:
Update Thing Client to fetch pending things


MHVLib/Clients/MHVThingClient.m
```diff
@@ -162,13 +182,48 @@ - (void)getThingsWithQueries:(MHVThingQueryCollection *)queries
          else
          {
              MHVThingQueryResults *queryResults = [self thingQueryResultsFromResponse:response];
```
namalu, line 184: I think that we need to check if the queryResults parsed correctly by verifying that it is not nil. If it is nil we should return an error so the caller knows that there is a problem with the response.

MHVLib/Clients/MHVThingClient.m
```diff
@@ -162,13 +182,48 @@ - (void)getThingsWithQueries:(MHVThingQueryCollection *)queries
...
              else
              {
-                 completion(nil, [NSError error:[NSError MHVUnknownError] withDescription:@"The GetThings results could not be extracted."]);
+                 if (results)
+                 {
+                     completion(results, nil);
+                 }
+                 else
+                 {
+                     completion(nil, [NSError error:[NSError MHVUnknownError] withDescription:@"The GetThings results could not be extracted."]);
```
namalu: I'm not quite sure what this error description means. We should be as detailed as possible with our error descriptions so debugging is easier for us and consumers of the SDK.

MHVLib/Types/MHVThingQueryResult.m
```diff
@@ -199,4 +199,33 @@ - (instancetype)init
...
+}
+
+- (void)mergeThingQueryResultCollection:(MHVThingQueryResultCollection *)collection
+{
+    for (MHVThingQueryResult *result in collection)
+    {
+        MHVThingQueryResult *existingResult = [self resultWithName:result.name];
+        if (existingResult)
+        {
+            [existingResult.things addObjectsFromArray:result.things.toArray];
```
namalu: We should use addObjectsFromCollection:

namalu (approved):
Just a few comments.

MHVLib/Clients/MHVThingClient.m
miburf: Added.

MHVLib/Clients/MHVThingClient.m
miburf: The other new one covers it, so removed; since could be no results

PR #65, opened 2017-06-02
Add support for uploading Blobs

miburf:
Add uploading of Blobs to ThingClient
Add image viewer to see downloaded blob images in SDKFeatures
Moved blob ThingClient tests to a new file, and added tests for uploads


MHVLib/Clients/MHVPersonClientProtocol.h
```diff
@@ -18,7 +18,7 @@
 
 #import "MHVClientProtocol.h"
 
-@class MHVApplicationSettings, MHVPersonInfo, MHVHealthRecordInfo, MHVGetAuthorizedPeopleSettings;
+@class MHVApplicationSettings, MHVPersonInfo, MHVHealthRecordInfo, MHVGetAuthorizedPeopleSettings, UIImage;
```
namalu: It looks like this is no longer needed.

MHVLib/Clients/MHVClientFactory.m
```diff
@@ -20,6 +20,7 @@
 #import "MHVPersonClient.h"
 #import "MHVPlatformClient.h"
 #import "MHVRemoteMonitoringClient.h"
+#import "MHVThingClient.h"
```
namalu, line 23: This appears to be added twice.

MHVLib/Clients/MHVThingClient.m
```diff
@@ -665,6 +666,94 @@ - (void)getPersonalImageWithRecordId:(NSUUID *)recordId
...
+    MHVASSERT_PARAMETER(completion);
+    
+    if (!completion)
+    {
+        return;
+    }
+    
+    if (!blobSource || !toThing || !contentType || !recordId)
+    {
+        completion(nil, [NSError MVHInvalidParameter]);
```
namalu: Let's add an error description here. It can be as simple as 'One or more required parameters is nil.'

MHVLib/Errors/MHVErrorConstants.h
```diff
@@ -28,6 +28,7 @@ typedef NS_ENUM(NSUInteger, MHVErrorType)
     MHVErrorTypeIOError,
     MHVErrorTypeUnauthorized,
     MHVErrorTypeOperationCancelled,
+    MHVErrorTypeNoResults,
```
namalu: It appears that this is not used anywhere - We should delete it.

MHVLib/Methods/MHVBlobUploadRequest.h
```diff
@@ -0,0 +1,49 @@
...
+@property (nonatomic, strong, readonly) NSURL *destinationURL;
+
+/**
+ Http chunk size for the blob upload
+ */
+@property (nonatomic, assign, readonly) NSUInteger chunkSize;
+
+- (instancetype)initWithBlobSource:(id<MHVBlobSourceProtocol>)blobSource
+                    destinationURL:(NSURL *)destinationURL
+                         chunkSize:(NSUInteger)chunkSize;
```
namalu: You may want to add documentation to this initializer - The fact that passing Zero for the chanceSize parameter actually uses a default value is not clear without looking at the implementation.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                                              completion:^(MHVThing *_Nullable thing, NSError *_Nullable error) { }];
+                });
+                
+                let(method, ^{
+                    return (MHVMethod *)requestedServiceOperationStep1;
+                });
+                
+                it(@"should get things", ^
+                   {
+                       [[method.name should] equal:@"GetThings"];
```
namalu: In general any asynchronous method should use asynchronous expectations. (i.e. [[expectedFutureValue(foo) shoudEventually] ...]. We are testing an asynchronous method here and you are using your knowledge of the current implementation details to develop these tests and test the areas of code that happen to execute procedurally. If the underlying implementation changes, then there is a possibility we well have test failures even if the code changes are valid. Our tests should always pass as long as the end result fulfills the expectation correctly, as it would in a real world situation.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                it(@"should have correct info xml", ^
+                   {
+                       [[theValue([method.parameters containsString:@"<id>AllergyThingKey</id><format><section>core</section>"\
+                                   "<section>blobpayload</section><xml/></format></group></info>"]) should] beYes];
+                   });
+            });
+    
+    context(@"RefreshBlobUrlsForThing Errors", ^
+            {
+                beforeAll(^{
```
namalu: This should be 'beforeEach'. The tests within this context make the same call and should not contain any left over state from the previous test. I strongly recommend that **all** test variables be reset beforeEach test to ensure there is no left over states.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                    requestedServiceOperationStep2 = nil;
+                    
+                    [thingClient refreshBlobUrlsForThing:allergyThing
+                                                recordId:recordId
+                                              completion:^(MHVThing *_Nullable thing, NSError *_Nullable error) { }];
+                });
+                
+                let(method, ^{
+                    return (MHVMethod *)requestedServiceOperationStep1;
+                });
```
namalu: I don't think this let is actually necessary. The documentation states that let 'declares a local helper variable that is re-initialized before every it block in all enclosed contexts'. However, you are just returning the variable requestedServiceOperationStep1 which is set one time in the 'beforeAll' block above. You can just use the requestedServiceOperationStep1 variable.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+SPEC_BEGIN(MHVThingClientBlobTests)
+
+describe(@"MHVThingClient", ^
+{
+    __block id<MHVHttpServiceOperationProtocol> requestedServiceOperationStep1;
+    __block id<MHVHttpServiceOperationProtocol> requestedServiceOperationStep2;
+    __block id<MHVHttpServiceOperationProtocol> requestedServiceOperationStep3;
+    
+    __block MHVServiceResponse *serviceResponseForBeginPut;
+    __block MHVServiceResponse *serviceResponseForBlobUpload;
```
namalu: I really think that all variables need to be set to a default state (probably nil) before each test. You are using your knowledge of the current order of operations to design these tests, and only resetting some of these values depending on the test. If the implementation changes we might end up with false pass/fail results due to left over states from previous tests.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                       __block NSError *requestError;
+                       [thingClient refreshBlobUrlsForThing:nil
+                                                   recordId:recordId
+                                                 completion:^(MHVThing *_Nullable thing, NSError *_Nullable error)
+                        {
+                            requestError = error;
+                        }];
+                       
+                       [[requestError should] beNonNil];
+                       [[theValue(requestError.code) should] equal:@(MHVErrorTypeRequiredParameter)];
```
namalu: The tests in this context need to use asynchronous expectations also. The only reason they work now is because of the current implementation of the method.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                    requestedServiceOperationStep2 = nil;
+                    
+                    [thingClient refreshBlobUrlsForThings:[[MHVThingCollection alloc] initWithThings:@[allergyThing, fileThing]]
+                                                 recordId:recordId
+                                               completion:^(MHVThingCollection *_Nullable things, NSError *_Nullable error) { }];
+                });
+                
+                let(method, ^{
+                    return (MHVMethod *)requestedServiceOperationStep1;
+                });
```
namalu: Here too... the local method variable is probably unneeded because requestedServiceOperationStep1 is set one time in the beforeAll block.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                it(@"should have correct info xml", ^
+                   {
+                       [[theValue([method.parameters containsString:@"<id>AllergyThingKey</id><id>FileThingKey</id><format>"\
+                                   "<section>core</section><section>blobpayload</section><xml/></format></group></info>"]) should] beYes];
+                   });
+            });
+    
+    context(@"RefreshBlobUrlsForThingCollection Errors", ^
+            {
+                beforeAll(^{
```
namalu: This should be 'beforeEach'. The tests within this context make the same call and should not contain any left over state from the previous test. I strongly recommend that **all** test variables be reset beforeEach test to ensure there is no left over states.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                       __block NSError *requestError;
+                       [thingClient refreshBlobUrlsForThings:nil
+                                                    recordId:recordId
+                                                  completion:^(MHVThingCollection *_Nullable things, NSError *_Nullable error)
+                        {
+                            requestError = error;
+                        }];
+                       
+                       [[requestError should] beNonNil];
+                       [[theValue(requestError.code) should] equal:@(MHVErrorTypeRequiredParameter)];
```
namalu: The tests in this context need to use asynchronous expectations also. The only reason they work now is because of the current implementation of the method.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                   {
+                       __block NSError *requestError;
+                       [thingClient downloadBlobData:nil
+                                          completion:^(NSData * _Nullable data, NSError * _Nullable error)
+                        {
+                            requestError = error;
+                        }];
+                       
+                       [[requestError should] beNonNil];
+                       [[theValue(requestError.code) should] equal:@(MHVErrorTypeRequiredParameter)];
```
namalu: The tests in this context need to use asynchronous expectations also. The only reason they work now is because of the current implementation of the method.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
...
+                   });
+                
+            });
+
+    context(@"UploadBlob step 1", ^
```
namalu: We probably shouldn't have tests that specifically call out the implementation details of the service. We should test the final result of the call which is either success or error. We can assume that all requirements are met and all of the method calls are successful, the result to the caller is success and the output from the method can be validated. However, if errors occur along the way web should verify the output contains a useful error (an error that specifically calls out what went wrong) and that the 'thing' parameter is nil. 'Step 1' is only step one because of the current implementation - If this changes in the future these tests are broken.

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,502 @@
+//
+// MHVThingClientBlobTests.m
```
namalu, line 2: I think many of these tests are tied to closely with the current implementation. I don't think we should be testing 'Step 1', 'Step 2', or 'Step 3', but instead, we should be testing the output that the caller gets from a method. The output includes the results provided in the completion block and no more (thing and error). We should base each test on 'if this happens, than the caller gets this'. I see a lot of test that are validating internal states which ties the tests to the implementation that we have at this point in time. If a code contributor modifies the implementation the should not have to rewrite the tests - Instead they should be able to use the tests as a way to ensure they didn't break anything.

namalu (changes requested):
I would like to make sure that our tests are not coupled so tightly to the current implementation. Tests should pass as long as the result from the caller is expected.

MHVLibTests/Clients/MHVThingClientBlobTests.m
miburf: Updated all the tests to do the shouldEventually() and do check results instead of internal states.

MHVLib/Errors/NSError+MHVError.m
```diff
@@ -39,6 +39,15 @@ + (NSError *)MVHInvalidParameter:(NSString *)message
                                       }];
 }
 
++ (NSError *)MVHRequiredParameterIsNil
+{
+    return [NSError errorWithDomain:kMHVErrorDomain
+                               code:MHVErrorTypeRequiredParameter
```
namalu, line 45: I'm not seeing MHVErrorTypeRequiredParameter in the MHVErrorConstants file?

MHVLibTests/Clients/MHVThingClientBlobTests.m
```diff
@@ -0,0 +1,465 @@
...
+
+describe(@"MHVThingClient", ^
+{
+    __block id<MHVHttpServiceOperationProtocol> requestedServiceOperationStep1;
+    __block id<MHVHttpServiceOperationProtocol> requestedServiceOperationStep2;
+    __block id<MHVHttpServiceOperationProtocol> requestedServiceOperationStep3;
+    
+    __block MHVServiceResponse *serviceResponseForStep1;
+    __block MHVServiceResponse *serviceResponseForStep2;
+    __block MHVServiceResponse *serviceResponseForStep3;
```
namalu: Instead of naming these step1, 2, 3, can we make the names represent what they are? It's hard to tell what is going on.

namalu (approved):
We are missing tests for downloading blobs to files and downloading the personal image We should add these to bring the code coverage to over 80%.

MHVLib/Errors/NSError+MHVError.m
miburf: It's there for me, the first one.

PR #67, opened 2017-06-06
Fixes for retrying requests and token refreshes

miburf:
Fixes for retrying requests and token refreshes
Add tests for retrying

### DIFF
--- a/MHVLib/Connection/MHVConnection.m
+++ b/MHVLib/Connection/MHVConnection.m
@@ -453,7 +453,7 @@ static NSInteger kInternalServerError = 500;
         if (request.completion)
         {
             request.completion(nil, [NSError error:[NSError MHVUnauthorizedError]
-                                   withDescription:@"The refreshed session authorization is invalid or expired."]);
+                                   withDescription:@"The session authorization was refreshed, but the new authorization is invalid or expired."]);
         }
         return;
     }
--- a/MHVLib/Connection/MHVConnection.m
+++ b/MHVLib/Connection/MHVConnection.m
@@ -18,6 +18,7 @@
 
 #import "MHVCommon.h"
 #import "MHVConfiguration.h"
+#import "MHVConfigurationConstants.h"
 #import "MHVConnection.h"
 #import "MHVAuthSession.h"
 #import "MHVMethod.h"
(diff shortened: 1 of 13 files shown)


Samples/SDKFeatures/SDKFeatures/ViewControllers/ActionPlanTasks/MHVActionPlanTaskListViewController.m
```diff
@@ -113,15 +113,14 @@ - (void)viewDidLoad
     
     self.tableView.dataSource = self;
     self.tableView.delegate = self;
-    
-    [self loadActionPlanTasks];
-    [self loadActionPlan];
```
mikenevi, line 118: Thanks for fixing that :)

mikenevi (approved):

MHVLib/Connection/MHVConnection.m
```diff
@@ -402,6 +445,20 @@ - (void)refreshSessionCredentialWithCompletion:(void(^_Nullable)(NSError *_Nulla
...
 {
+    // If it already refreshed the token for this request, don't retry since refreshing again won't fix the problem
+    if (request.hasRefreshedToken)
+    {
+        MHVLOG(@"The refreshed session authorization is invalid or expired.");
+        
+        if (request.completion)
+        {
+            request.completion(nil, [NSError error:[NSError MHVUnauthorizedError]
+                                   withDescription:@"The refreshed session authorization is invalid or expired."]);
```
namalu: We should add to the description that an attempt to refresh the token was made but failed.

namalu (changes requested):
I'm not seeing any updated tests to verify the new functionality added.

MHVLibTests/Connection/MHVConnectionTests.m
```diff
@@ -412,16 +398,174 @@ @interface MHVConnection (Testing)
...
+                    
+                    MHVMethod *method = [MHVMethod getThings];
+                    method.parameters = @"GETTHINGSBODY";
+                    [testConnection executeHttpServiceOperation:method
+                                                     completion:^(MHVServiceResponse * _Nullable response, NSError * _Nullable error) { }];
+                });
+                
+                it(@"requestCount should be 3", ^
+                   {
+                       [[expectFutureValue(requestCount) shouldEventuallyBeforeTimingOutAfter(10)] equal:@(3)];
```
namalu, line 426: I think we also need to make sure that requests don't end up in an infinite loop. This test will succeed once the requestCount variable == 3, but if we ned to make sure it never reaches 4.

namalu (approved):

MHVLibTests/Connection/MHVConnectionTests.m
miburf: Added a test that it doesn't reach 4

MHVLib/Connection/MHVConnection.m
miburf: Improved the description

PR #68, opened 2017-06-06
Adding caching mechanism to ServiceOperation objects

anbutle:
Allows clients to associate a cache with any ServiceOperation. The MHVConnection handles reading/writing to and from the cache based off the cacheKey (which can be defined in the ServiceOperation).


MHVLib/Clients/MHVVocabularyClientProtocol.h
```diff
@@ -111,6 +111,19 @@ NS_ASSUME_NONNULL_BEGIN
                              maxResults:(NSNumber *_Nullable)maxResults
                              completion:(void(^)(MHVVocabularyCodeSetCollection *_Nullable codeSet, NSError *_Nullable error))completion;
 
+
+/**
+ * Enables or disables caching of vocabulary responses
+ *
+ * @param cacheEnabled If YES, caching will be enabled for all subsequent calls. If NO, caching will be disabled for all subsequent calls.
```
namalu: We should add a note that the default value is YES.

MHVLib/Connection/MHVConnection.m
```diff
@@ -255,13 +266,19 @@ - (void)executeMethodRequest:(MHVHttpServiceRequest *)request
         }
         else
         {
+            if (request.serviceOperation.cache)
+            {
+                [request.serviceOperation.cache setObject:response forKey:cacheKey];
+            }
```
namalu: I think this is done in the wrong place - I believe you cannot save the response until after it is parsed and we determine there are no errors. For XML responses, everything is a 200 SUCCESS, and there is a status code in the payload that may or may not have an error.

MHVLib/Connection/MHVConnection.m.orig
```diff
@@ -0,0 +1,574 @@
+//
+//  MHVConnection.m
```
namalu: You should delete this .orig file.

MHVLib/Methods/MHVBlobDownloadRequest.m
```diff
@@ -27,4 +29,10 @@ - (instancetype)initWithURL:(NSURL *)url
     return self;
 }
 
+- (NSString *)getCacheKey
+{
+    [NSException throwNotImpl];
```
namalu: Whoa there cowboy!!! Let's not throw an exception here, lets just log it for debugging purposes.

namalu (approved):
There is an important comment about where you are saving the response - Please address this before merging!

PR #69, opened 2017-06-06
Initial Medication Tracker

kaylaerdmann:


Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -0,0 +1,156 @@
+//
+//  AddMedicationViewController.swift
+//  MedicationTracker
+//
+//  Created by Kayla Davis on 6/1/17.
+//  Copyright © 2017 Kayla Davis. All rights reserved.
+//
```
miburf: All the files should have the Microsoft copyright which you can copy from any of the MHV files.

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+
+import UIKit
+
+class AddMedicationViewController: UIViewController, UITextFieldDelegate,  UIPickerViewDelegate, UIPickerViewDataSource, UITableViewDelegate, UITableViewDataSource {
```
miburf: It’s a style thing, but all our code has more returns :)  Before every { or }.  It took me a bit to get used to it at the beginning, since my style was like what you’re doing.  But it’s easier to read if everything looks the same.

    func isEmpty() -> Bool
    {
        self.layer.borderWidth = 1.0
        if self.text!.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines).isEmpty
        {
            self.layer.borderColor = UIColor.red.cgColor
            errorLabel?.isHidden = false
            return true
        }
        else
        {
            self.layer.borderColor = UIColor.clear.cgColor
            errorLabel?.isHidden = true
            return false
        }
    }

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+    
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let selectedCell: UITableViewCell = tableView.cellForRow(at: indexPath)!
+        
+        nameField.text = selectedCell.textLabel!.text!
+        nameTableView.isHidden = true
+    }
+
+    
+    // MARK: UIPickerView Delegation
```
miburf: Creating a subclass of UIMedicationTextField (or another UITextField subclass) -> UIPickerTextField would simplify.  It can have all the UIPickerView delegate methods there, then don't need it in the view controllers, and don't have to do the checks here for which picker data to show.

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+        
+        doseUnitField.inputView = dosePicker
+        dosePicker.delegate = self
+        nameField.errorLabel = medicationErrorLabel
+        doseAmountField.errorLabel = doseAmountErrorLabel
+        
+        nameTableView.delegate = self
+        nameTableView.dataSource = self
+        nameTableView.isScrollEnabled = true
+        nameTableView.isHidden = true
```
miburf: nit, you should be able to set all these in the interface builder

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+        textField.resignFirstResponder()
+        return true
+    }
+    
+    func textFieldShouldEndEditing(_ textField: UITextField) -> Bool {
+        let text = textField as! UIMedicationTextField
+        _ = text.isEmpty()
+        if (textField == doseAmountField){
+            _ = text.isNumeric()
+        }
```
miburf: You could add a var bool for "shouldBeNumeric" to your UIMedicationTextField; and if you check out @IBInsectable that would let it be set from interface builder.  
@IBInspectable var shouldBeNumeric:Bool = false

Then this method could call text.isValid()  and the textfield can have a if (self.shouldBeNumeric) {} check.  

That would simplify especially on the forms with more fields.

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+        nameTableView.isHidden = true
+
+        return true
+    }
+    
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        let substring = (textField.text! as NSString).replacingCharacters(in: range, with: string)
+        if substring.characters.count >= minSearchSize {
+            nameTableView.isHidden = false
```
miburf: nit, I'd move this down to inside the DispatchQueue below, and could then do nameTableView.isHidden = autocompleteContents.count == 0

That way it doesn't show before results have arrived, and doesn't show if there aren't any results.

miburf: There also isn't a check for which field is being edited, so typing 1111 in the Dosage field shows the autocomplete for the name :)

Samples/MedicationsTracker/MedicationsTracker/MedicationEntryUtils.swift
```diff
@@ -0,0 +1,36 @@
...
+                         "Unit (unt)","Units per milliliter (unt/ml)"]
+}
+
+class MedicationVocabSearcher {
+    let minSearchSize = 3
+
+    func searchForMeds(searchValue: String, completion: @escaping(MHVVocabularyCodeItemCollection?) -> Void){
+        let connection = MHVConnectionFactory.current().getOrCreateSodaConnection(with: HVFeaturesConfiguration.configuration())
+        let key = MHVVocabularyKey.init(name: "RxNorm Active Medicines", andFamily: "RxNorm", andVersion: "09AB_091102F", andCode: nil)
+        connection.vocabularyClient()?.searchVocabulary(withSearchValue: searchValue, searchMode: MHVSearchMode.contains, vocabularyKey: key!, maxResults: 25, completion: { (matchedMeds: MHVVocabularyCodeSetCollection?, error: Error?) in
```
miburf: Not something to handle yet, but to be aware and add a TODO:
It could be typing ABCD that the results arrive out of order, so those for ABCD arrive, and then the results for ABC.  An autocomplete in the old Band app used the date of the request, so it wouldn't show the results from an earlier date than that it was displaying.

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+    // MARK: Actions
+    @IBAction func addMedication(_ sender: UIButton) {
+        if(fieldsAreComplete()){
+            let medication = medicationBuilder!
+                .begin(mhvThing: medicationThing!)
+                .setName(name: nameField.text!)
+                .setStrengthIfNotNil(amount: strengthAmountField.text!, unit: strengthUnitField.text!)
+                .setDoseIfNotNil(amount: doseAmountField.text!, unit: doseUnitField.text!)
+                .setFrequencyIfNotNil(frequency: howOftenField.text!)
+                .constructMedication()
```
miburf: This chaining works, but makes it really hard for showing errors.  Especially if different pages (add vs edit) need to show a different error for a step, or something different should happen depending on the step that failed.

Returning a Bool if the method succeeds or not would let you handle errors, for example to show an alert and put the focus on the field with the error:

if !medication.updateName(nameField.text) {
  Alert("Name is required")
  nameField.becomeFirstResponder()
  return;
}

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+            let connection = MHVConnectionFactory.current().getOrCreateSodaConnection(with: HVFeaturesConfiguration.configuration())
+            connection.thingClient()?.update(medication, record: connection.personInfo!.selectedRecordID, completion: { (error: Error?) in
+            })
+        }
+    }
+    
+    // MARK: Helper Functions
+    func fieldsAreComplete() -> Bool {
+        if(nameField.isEmpty() || !doseAmountField.isNumeric() || !strengthAmountField.isNumeric()){
```
miburf: nit, return !nameField.isEmpty() && doseAmountField.isNumeric() && strengthAmountField.isNumeric()
makes it shorter :)

Samples/MedicationsTracker/MedicationsTracker/MedicationBuilder.swift
```diff
@@ -0,0 +1,53 @@
...
+    var thing: MHVThing?
+    var med: MHVMedication?
+    
+    func begin(mhvThing: MHVThing) -> MedicationBuilder{
+        thing = mhvThing
+        med = mhvThing.medication()
+        return self
+    }
+    
+    func setName(name: String) -> MedicationBuilder {
```
miburf: See earlier about returning Bool if these succeed or fail.

In ObjectiveC, set is a special prefix for a setter and those typically don't return a value, to be safe, maybe updateName() instead

Samples/MedicationsTracker/MedicationsTracker/UIMedicationTextField.swift
```diff
@@ -0,0 +1,46 @@
...
+                errorLabel?.isHidden = false
+                return false
+            }
+        }
+        
+        self.layer.borderColor = UIColor.clear.cgColor
+        errorLabel?.isHidden = true
+        return true
+    }
+
```
miburf: nit, don't have to change now, but so you wouldn't have to update multiple places when the design changes (since it inevitably will), I'd do 2 methods:  showNormalStyle()  showErrorStyle()

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+    @IBOutlet weak var doseUnitErrorLabel: UILabel!
+    @IBOutlet weak var howOftenErrorLabel: UILabel!
+    @IBOutlet weak var nameTableView: UITableView!
+    
+    //Mark: Initializers
+    init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?, medication: MHVThing, builder: MedicationBuilder, searcher: MedicationVocabSearcher) {
+        super.init(nibName: nibNameOrNil, bundle: nibBundleOrNil)
+        medicationThing = medication
+        medicationBuilder = builder
+        self.searcher = searcher
```
miburf: nit, i'd do either property or self.property, but not mix in one method

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+                .constructMedication()
+            let connection = MHVConnectionFactory.current().getOrCreateSodaConnection(with: HVFeaturesConfiguration.configuration())
+            connection.thingClient()?.createNewThing(medication, record: connection.personInfo!.selectedRecordID, completion: { (error: Error?) in
+            })
+        }
+    }
+    
+    // MARK: Helper Functions
+    func fieldsAreComplete() -> Bool {
+        if(nameField.isEmpty() || !doseAmountField.isNumeric()){
```
miburf: nit, return !nameField.isEmpty() && doseAmountField.isNumeric() is shorter

kaylaerdmann:
I'm looking for feedback on the design of this part of the system. Things that I know need to be changed: 
* the navigation
* the flow of the pages
* ui

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+    
+    func pickerView( _ pickerView: UIPickerView, titleForRow row: Int, forComponent component: Int) -> String? {
+        return dosePickerData[row]
+    }
+    
+    func pickerView( _ pickerView: UIPickerView, didSelectRow row: Int, inComponent component: Int) {
+        doseUnitField.text = dosePickerData[row]
+    }
+    
+    // MARK UITableView
```
miburf: Similarly to the UIPickerTextField, could have a UIAutocompleteTextField that handled the table; then it's in one place too.  There's only one per view now, so wouldn't need to do it yet; but would be helpful if more fields with autocomplete are added.

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.xib
```diff
@@ -0,0 +1,107 @@
...
+                <outlet property="doseAmountField" destination="zDb-Pa-aaz" id="bpJ-XX-XPN"/>
+                <outlet property="doseUnitField" destination="pGL-ah-QHf" id="FOt-Mn-lPw"/>
+                <outlet property="medicationErrorLabel" destination="HGf-QJ-xas" id="7da-3D-XmA"/>
+                <outlet property="nameField" destination="mmL-qi-kM0" id="WPh-I6-B57"/>
+                <outlet property="nameTableView" destination="QS1-Ce-2wC" id="CX0-iv-8ci"/>
+                <outlet property="view" destination="i5M-Pr-FkT" id="sfx-zR-JGt"/>
+            </connections>
+        </placeholder>
+        <placeholder placeholderIdentifier="IBFirstResponder" id="-2" customClass="UIResponder"/>
+        <view clearsContextBeforeDrawing="NO" contentMode="scaleToFill" id="i5M-Pr-FkT">
```
miburf, line 24: When you're ready, I'll show you about AutoLayout.  As it is now, if you run the app on a smaller phone, the fields can be too wide & be hidden by the keyboard.

Dismissing the keyboard is tricky, especially the Dosage number one because there isn't a Return key on the numeric keypad.  If you look in the Insights app code, KHRKeyboardAccessoryView adds the next/previous/done button above the keyboard.  It's handy to not have to do other weird stuff like detecting taps on the background.  You can try using as-is in Objective-C to start, everything looks like it will work except the styling it does.

miburf (changes requested):
Looks good, some comments...

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
kaylaerdmann: done

kayladavis: done

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
kaylaerdmann: no longer applicable

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
kaylaerdmann: no longer applicable

Samples/MedicationsTracker/MedicationsTracker/UIMedicationTextField.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
kayladavis: done

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/MedicationBuilder.swift
kaylaerdmann: done

Samples/MedicationsTracker/MedicationsTracker/MedicationEntryUtils.swift
kaylaerdmann: TODO added

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -2,155 +2,98 @@
...
     var medicationBuilder: MedicationBuilder?
-    let dosePicker = UIPickerView()
-    var dosePickerData = HVUnitTypes.doseUnits
-    var autoComplete: MHVVocabularyCodeItemCollection?
-    var searcher: MedicationVocabSearcher?
-    let minSearchSize = 3
     
     //MARK: UI Properties
-    @IBOutlet weak var nameField: UIMedicationTextField!
+    @IBOutlet weak var nameField: UIAutocompleteTextField!
```
miburf: nit, don't need the prefix for swift, especially UI as that's reserved by Apple

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -2,155 +2,98 @@
...
+    {
         super.viewDidLoad()
         
-        doseUnitField.inputView = dosePicker
-        dosePicker.delegate = self
+        // Setup custom field types
+        nameField.setup()
+        doseUnitField.setup(pickerData: HVUnitTypes.doseUnits)
+        
+        // Set error lables where needed
```
miburf: nit, typo

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.swift
```diff
@@ -2,155 +2,98 @@
...
-            return Int(self.autoComplete!.count())
-        } else{
-            return 0
-        }
-    }
-    
-    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
-        let selectedCell: UITableViewCell = tableView.cellForRow(at: indexPath)!
+        // Hide autocomplete table when done editing
+        nameField.tableView?.isHidden = true
```
miburf: I think you could do this in the AutocompleteTextField too.

override func resignFirstResponder() -> Bool
{
self.tableView?.isHidden = true
return super.resignFirstResponder()
}

Samples/MedicationsTracker/MedicationsTracker/EditMedicationViewController.swift
```diff
@@ -2,221 +2,166 @@
...
-        strengthUnitField.inputView = strengthPicker
-        dosePicker.delegate = self
-        strengthPicker.delegate = self
+        
+        // Setup custom field types
+        nameField.setup()
+        doseUnitField.setup(pickerData: HVUnitTypes.doseUnits)
+        strengthUnitField.setup(pickerData: HVUnitTypes.strengthUnits)
+        
+        // Set error lables where needed
```
miburf: :) copied typo

Samples/MedicationsTracker/MedicationsTracker/UIAutocompleteTextField.swift
```diff
@@ -0,0 +1,91 @@
...
+//
+
+import UIKit
+
+class UIAutocompleteTextField: UIMedicationTextField, UITableViewDelegate, UITableViewDataSource, UITextFieldDelegate
+{
+    // MARK: Properties
+    var autoComplete: MHVVocabularyCodeItemCollection?
+    var tableView: UITableView?
+    var tableHight: CGFloat = 44 * 4
```
miburf: nit, typo

Samples/MedicationsTracker/MedicationsTracker/AddMedicationViewController.xib
kaylaerdmann: kicking this to a later pr

miburf (approved):

PR #72, opened 2017-06-08
Personal/namalu/implement platform client

namalu:
Implementation and testing for MHVPlatformClient


MHVLib/Clients/MHVPlatformClient.m
```diff
@@ -115,39 +133,87 @@ - (void)newApplicationCreationInfoWithCompletion:(void(^)(MHVApplicationCreation
...
+                    if ([typeId caseInsensitiveCompare:definition.typeId.UUIDString] == NSOrderedSame)
+                    {
+                        [definitions setObject:definition forKey:typeId];
+                        
+                        [typeDefinitions removeObject:definition];
+                        
+                        break;
+                    }
+                }
+            }
```
miburf, line 202: It should log if there's some mismatch, typeDefinitions.count != 0 at the end.  Using [typeIds contains:definition.typeId.UUIDString] would simplify to only need one loop here

MHVLib/ThingTypes/Definition/MHVThingTypeDefinitionRequestParameters.m
```diff
@@ -0,0 +1,97 @@
...
+        [writer writeElement:c_element_section value:@"effectivedatexpath"];
+    }
+    
+    // Write each image type as a separate element
+    for (NSString *imageType in self.imageTypes)
+    {
+        [writer writeElement:c_element_image_type value:imageType];
+    }
+    
+    // Write the last clietn refresh date (if not nil)
```
miburf: *client

miburf (approved):

MHVLib/Clients/MHVPlatformClient.m
namalu: Unfortunately, the casing of the UUID string is not guaranteed to be uppercase or lowercase, so I need to do a case insensitive comparison.

PR #77, opened 2017-06-10
Adding additional things and types to the sdk

anbutle:
Adding a number of types and things which were missing from the sdk. Also adding tests to validate the parsing of these types.


MHVLib/Types/MHVPlanObjective.h
```diff
@@ -26,10 +26,16 @@
 @property (readwrite, nonatomic, strong) MHVStringNZNW *name;
 @property (readwrite, nonatomic, strong) MHVStringNZNW *descriptionText;
 @property (readwrite, nonatomic, strong) NSString *state;
-@property (readwrite, nonatomic, strong) MHVPlanOutcomeCollection *outcomes;
+@property (readwrite, nonatomic, strong) MHVPlanOutcomeList *outcomes;
```
namalu: Is this 'plan' as in 'Action Plan'? If so, are we supposed to be accessing action plan data using the REST services?

namalu (approved):

MHVLibTests/ThingTypes/MHVBodyDimensionTests.m
```diff
@@ -0,0 +1,59 @@
...
+             
+             context(@"Deserialize", ^
+                     {
+                         it(@"should deserialize correctly", ^
+                            {
+                                MHVBodyDimension *bodyDimension = (MHVBodyDimension*)[XReader newFromString:objectDefinition withRoot:[MHVBodyDimension XRootElement] asClass:[MHVBodyDimension class]];
+                                
+                                [[bodyDimension.when.description should] equal:@"01/01/05 06:00 AM"];
+                                [[bodyDimension.measurementName.description should] equal:@"Left bicep size"];
+                                [[theValue(bodyDimension.value.display.value) should] equal:theValue(0.14999999999999999)];
```
namalu: For testing floating point values you should probably use [[theValue(bodyDimension.value.display.value) should] beBetween:theValue(14.999) and:theValue(15.001)] - That is, if it is acceptable to have floating point errors on deserialization.

namalu (approved):

MHVLib/Types/MHVPlanObjective.h
anbutle: I believe this is the object used for actions plans, but it is not called such in the HV docs. Even if we support it with the REST services, we should have it in the core HV code I thought?

PR #78, opened 2017-06-12
Add setting personal image and delete a few old files

miburf:
Starting to cleanup and delete all the old code.
Add setting personal image to Thing client and delete a few old files


MHVLib/Clients/MHVThingClient.m
```diff
@@ -666,6 +667,57 @@ - (void)getPersonalImageWithRecordId:(NSUUID *)recordId
...
+- (void)setPersonalImage:(NSData *)imageData
+             contentType:(NSString *)contentType
+                recordId:(NSUUID *)recordId
+              completion:(void (^_Nullable)(NSError *_Nullable error))completion
+{
+    MHVASSERT_PARAMETER(imageData);
+    MHVASSERT_PARAMETER(contentType);
+    MHVASSERT_PARAMETER(recordId);
+    
+    if (!imageData || !contentType || !recordId)
```
namalu, line 679: Is it okay for contentType to be an empty string?

namalu (approved):

MHVLib/Clients/MHVThingClient.m
miburf: I think it does need to be set so browsers know how to decode the data.  I don't think other platforms all have the nice imageWithData: that iOS does that converts anything so may look at the type to do a JPG vs PNG decoder.  And really should only be image/* types so it can be displayed.  I'll add a check for the standard http image content types.

PR #81, opened 2017-06-12
Dev/mikenev/s67/preview2

mikenevi:


MHVLib/Models/MHVModelBase.m
```diff
@@ -197,36 +197,44 @@ + (void)setProperties:(NSObject<MHVDataModelProtocol> *)object fromDictionary:(N
...
-                        
-                        //All property type classes must implement, objectParameters ignored except for arrays and dates
-                        value = [[propertyClass alloc] initWithObject:dictionary[propertyNameInDictionary]
-                                                     objectParameters:objectParameters];
-                        if (value)
+                        else
                         {
-                            [object setValue:value forKey:propertyName];
+                            // for base types (int, long, etc)
+                            [object setValue:dictionary[propertyNameInDictionary] forKey:propertyName];
```
mikenevi, line 237: This one is interesting. Without this change, the json deserializer doesn't deserialize to base objects like longs. That's a problem for MHVTime where the minutes and hours are stored as ints. This approach made the most sense to me. Otherwise, we could change the MHVTime to use NSNumber, though that would be a more far-reaching change.

namalu (approved):

PR #84, opened 2017-06-13
Update REST API calls that do not get a response object to no longer …

mikenevi:
…expect one.

### DIFF
--- a/MHVLib/Clients/MHVRemoteMonitoringClient.m
+++ b/MHVLib/Clients/MHVRemoteMonitoringClient.m
@@ -48,7 +48,7 @@
     return self;
 }
 
-- (void)requestWithPathAndResultClass:(NSString *_Nonnull)path
+- (void)requestWithPath:(NSString *_Nonnull)path
              httpMethod:(NSString *_Nonnull)httpMethod
              pathParams:(NSDictionary<NSString *, NSString *> *_Nullable)pathParams
             queryParams:(NSDictionary<NSString *, NSString *> *_Nullable)queryParams
--- a/MHVLib/Clients/MHVRemoteMonitoringClient.m
+++ b/MHVLib/Clients/MHVRemoteMonitoringClient.m
@@ -48,13 +48,12 @@
     return self;
 }
 
-- (void)requestWithPath:(NSString *_Nonnull)path
+- (void)requestWithPathAndResultClass:(NSString *_Nonnull)path
              httpMethod:(NSString *_Nonnull)httpMethod
              pathParams:(NSDictionary<NSString *, NSString *> *_Nullable)pathParams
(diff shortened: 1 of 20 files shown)


MHVLib/Clients/MHVRemoteMonitoringClient.m
```diff
@@ -48,13 +48,12 @@ - (instancetype)initWithConnection:(id<MHVConnectionProtocol>)connection
     return self;
 }
 
-- (void)requestWithPath:(NSString *_Nonnull)path
+- (void)requestWithPathAndResultClass:(NSString *_Nonnull)path
```
miburf: I wouldn't include the AndResultClass here, just the resultClass change below

MHVLib/Clients/MHVRemoteMonitoringClient.m
```diff
@@ -104,4 +102,53 @@ - (void)requestWithPath:(NSString *_Nonnull)path
...
+    {
+        NSMutableString *queryPath = [path mutableCopy];
+        
+        [pathParams enumerateKeysAndObjectsUsingBlock:^(id _Nonnull key, id _Nonnull obj, BOOL *_Nonnull stop)
+         {
+             [queryPath replaceCharactersInRange:[queryPath rangeOfString:[NSString stringWithFormat:@"{%@}", key]] withString:obj];
+         }];
+        
+        path = queryPath;
+    }
```
miburf: Both methods here do this, I'd make it into a updatePath:withPathParams: method they can share

miburf: Both methods here do this, I'd make it into a updatePath:withPathParams: method they can share.

MHVLib/Models/Generated/Categories/MHVActionPlanTasksApi.h
```diff
@@ -43,11 +43,10 @@ NS_ASSUME_NONNULL_BEGIN
 ///
 /// @param actionPlanTask 
 /// 
-///  code:200 message:"OK",
 ///  code:201 message:"Created",
 ///  code:0 message:"Error"
 ///
-/// @return NSObject*
+/// @return MHVActionPlanTaskInstanceV2*
```
miburf, line 49: I know it's the generated code from that big template, but they really return void & have the completion that returns the result object or error.

miburf (approved):

MHVLib/Models/Generated/Categories/MHVActionPlanTasksApi.h
mikenevi: You're right, I'll make a note of that and maybe fix it later. It is generated code.

PR #85, opened 2017-06-14
Initial action plan creation

kaylaerdmann:
With some stubbed out methods for the action plan wizard


Samples/MedicationsTracker/MedicationsTracker/ActionPlan.swift
```diff
@@ -0,0 +1,94 @@
...
+
+import Foundation
+
+class ActionPlan
+{
+    var actionPlanInstance: MHVActionPlanInstanceV2?
+    
+    func getOrCreateActionPlan()
+    {
+        let connection = MHVConnectionFactory.current().getOrCreateSodaConnection(
```
miburf: an init(connection) initializer will make it easier for testing later, makes it simple to pass in a mock connection

Samples/MedicationsTracker/MedicationsTracker/ActionPlan.swift
```diff
@@ -0,0 +1,94 @@
...
+        
+        // Get action plan
+        findActionPlan(connection: connection, completion:
+            {
+                (actionPlan) in self.actionPlanInstance = actionPlan
+            })
+        
+        // If the action plan is nil create one
+        if self.actionPlanInstance == nil
+        {
```
miburf: this if self.actionPlanInstance == nil will need to be inside the findActionPlan completion.  A createMedicationPlan(completion: @escaping(MHVActionPlanTaskInstanceV2?)) for it to call if findActionPlan doesn't find one works-- and the same completion can be passed along.

Samples/MedicationsTracker/MedicationsTracker/ActionPlan.swift
```diff
@@ -0,0 +1,94 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+class ActionPlan
+{
+    var actionPlanInstance: MHVActionPlanInstanceV2?
+    
+    func getOrCreateActionPlan()
```
miburf: This looks like it will need a completion as well, so nothing tries to use self.actionPlanInstance before it's set

Samples/MedicationsTracker/MedicationsTracker/ActionPlan.swift
```diff
@@ -0,0 +1,94 @@
...
+    
+    
+    func createAndAttachAssociatedTask(task: MHVActionPlanTaskV2, completion: @escaping(MHVActionPlanTaskInstanceV2?) -> Void)
+    {
+        let connection = MHVConnectionFactory.current().getOrCreateSodaConnection(
+            with: HVFeaturesConfiguration.configuration())
+        
+        connection.remoteMonitoringClient()?.actionPlanTasksCreate(withActionPlanTask: task, completion:
+            { (taskInstance: MHVActionPlanTaskInstanceV2?, error: Error?) in
+                self.actionPlanInstance?.associatedTasks?.append(taskInstance!)
```
miburf: taskInstance can be nil if there's an error

Samples/MedicationsTracker/MedicationsTracker/ActionPlan.swift
```diff
@@ -0,0 +1,94 @@
...
+            let objective = MHVObjective.init()
+            objective.name = "Take your medication"
+            objective.outcomeName = "Medication trends per week"
+            objective.outcomeType = "Other"
+            actionPlan.objectives = [objective]
+            
+            connection.remoteMonitoringClient()?.actionPlansCreate(withActionPlan: actionPlan, completion:
+                {
+                    (actionPlanInstance: MHVActionPlanInstanceV2?, error: Error?) in
+                        self.actionPlanInstance = actionPlanInstance
```
miburf: Your completion on getOrCreateActionPlan & the methods that call it will need to handle not having a plan, and the actionPlansCreate failing.  (Phone in airplane mode would do that)

Samples/MedicationsTracker/MedicationsTracker/ActionPlan.swift
```diff
@@ -0,0 +1,94 @@
...
+    private func findActionPlan(connection: MHVSodaConnectionProtocol, completion: @escaping(MHVActionPlanInstanceV2?) -> Void)
+    {
+        let connection = MHVConnectionFactory.current().getOrCreateSodaConnection(
+            with: HVFeaturesConfiguration.configuration())
+        
+        // Iterate though current action plans, until you find one that matches the app id
+        connection.remoteMonitoringClient()?.actionPlansGet(withMaxPageSize: 1000, completion:
+            { (actionPlans: MHVActionPlansResponseActionPlanInstanceV2_?, error: Error?) in
+                let plans = actionPlans?.plans as! [MHVActionPlanInstanceV2]
+                for actionPlan in plans
```
miburf: I'm not sure in Swift what will happen here if there's an error so actionPlans is nil.
_if let plans = actionPlans?.plans as? [MHVActionPlanInstanceV2]_  I think would handle

Samples/MedicationsTracker/MedicationsTracker/ActionPlanTaskBuilder.swift
```diff
@@ -0,0 +1,96 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import Foundation
+
+class ActionPlanTaskBuilder
+{
+    private(set) var actionPlanTask: MHVActionPlanTaskV2?
+    
+    func buildActionPlanTask(med: MHVMedication) -> ActionPlanTaskBuilder
```
miburf: return the constructed task _-> MHVActionPlanTaskV2_ or Bool for success makes more sense (and matches the method name)

Samples/MedicationsTracker/MedicationsTracker/ActionPlanTaskBuilder.swift
```diff
@@ -0,0 +1,96 @@
...
+        actionPlanTask!.imageUrl = "http://img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE13a3S?ver=884e"
+        actionPlanTask!.thumbnailImageUrl = "http://img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE12EQP?ver=cba6"
+        actionPlanTask!.trackingPolicy = createTrackingPolicy(med.name.text)
+        
+        return self
+    }
+    
+    func updateFrequencyMetric(windowType: String) -> Bool
+    {
+        let freqMetric = MHVActionPlanFrequencyTaskCompletionMetricsV2.init()
```
miburf: the update methods can all use _guard actionPlanTask_ to make sure the buildActionPlanTask happened.
OR
could you make it _private(set) var actionPlanTask: MHVActionPlanTaskV2.init()_  and then the buildActionPlanTask can become updateMedication() ?

Samples/MedicationsTracker/MedicationsTracker/HVUnitTypes.swift
```diff
@@ -20,18 +20,54 @@ import Foundation
...
+    case None = "None"
+    case Daily = "Daily"
+    case Weekly = "Weekly"
+}
+
+enum ReminderState: String
+{
+    case Unknown = "Unknown"
+    case Off = "Off"
+    case Ontime = "OnTime"
```
miburf: nit, OnTime

Samples/MedicationsTracker/MedicationsTracker/HVUnitTypes.swift
```diff
@@ -20,18 +20,54 @@ import Foundation
...
+    case FiveMin = "Before5Minutes"
+    case TenMin = "Before10Minutes"
+    case FiveteenMin = "Before15Minutes"
```
miburf: Fiveteen :)  Must be how it's spelled in Canadian

Samples/MedicationsTracker/MedicationsTracker/Transitioner.swift
```diff
@@ -0,0 +1,41 @@
...
+struct Transitioner
+{
+    
+    /*
+     * Handles the transition logic for the stored view controller
+     */
+    func transition()
+    {
+        
+    }
```
miburf: I think you'll need some changes when you get to these, at least passing in the navigationController for pushing the next page view controller.  (But no need to do  that until you get to creating the views and navigating between)

Samples/MedicationsTracker/MedicationsTracker/ActionPlan.swift
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    if actionPlan.organizationId == HVFeaturesConfiguration.configuration().masterApplicationId.uuidString
```
miburf: Swift may not optimize this, so could be building configuration and converting UUIDs for every plan in the loop.  You should make a local variable for it to avoid

Samples/MedicationsTracker/MedicationsTracker/ActionPlanTaskBuilder.swift
```diff
@@ -0,0 +1,96 @@
...
+        let schedule = MHVScheduleV2.init()
+        schedule.reminderState = reminderState
+        schedule.scheduledDays = scheduledDays
+        schedule.scheduledTime = time
+        
+        actionPlanTask!.schedules =  schedules + [schedule]
+        return true
+    }
+    
+    func constructActionPlanTask() -> MHVActionPlanTaskV2
```
miburf: Is there any validation to do?  in case schedule didn't get set, etc.

miburf (changes requested):

PR #88, opened 2017-06-15
Change from UIWebView to WKWebView

miburf:
Change from UIWebView to WKWebView
Move logic to AuthDelegate
Add AuthDelegate tests


Example/Tests/Controllers/MHVBrowserAuthBrokerTests.m
```diff
@@ -0,0 +1,348 @@
...
+@property (nonatomic, strong) MHVBrowserController *browserController;
+
+@end
+
+
+#pragma mark - Tests
+
+SPEC_BEGIN(MHVBrowserAuthBrokerTests)
+
+describe(@"MHVBrowserAuthBrokerTests", ^
```
namalu: I think this describe should actually just be MHVBrowserAuthBroker.

Example/Tests/Controllers/MHVBrowserAuthBrokerTests.m
```diff
@@ -0,0 +1,348 @@
...
+        authBroker.browserController = [[MHVTestBrowserController alloc] initWithAuthBroker:authBroker];
+        
+        presentedViewControllerAction = nil;
+        presentedViewController = nil;
+        returnedSuccessUrl = nil;
+        returnedError = nil;
+        returnedNavigationPolicy = -1;
+    });
+    
+    context(@"authenticateWithViewController", ^{
```
namalu: When doing BDD the describe, context, and it statements should form a complete sentence e.g "when authenticateWithViewController is called". This is important for making failures clear to someone who contributes and has no knowledge of our test system and limited knowledge of our codebase. If a test fails, the contributor will see "MHVBrowserAuthBroker_whenAuthenticateWithViewControllerIsCalled_shouldPresentTheProvidedViewController." The usage of English-like sentences is a core principle of BDD.

healthvault-ios-sdk/Classes/Controllers/MHVBrowserAuthBroker.h
```diff
@@ -18,7 +18,8 @@
 
 #import "MHVBrowserController.h"
 #import "MHVBrowserAuthBrokerProtocol.h"
+#import <WebKit/WebKit.h>
 
-@interface MHVBrowserAuthBroker : MHVBrowserController<MHVBrowserAuthBrokerProtocol>
+@interface MHVBrowserAuthBroker : NSObject <MHVBrowserAuthBrokerProtocol>
```
namalu: We should probably move this out of the Controllers directory and put it in the Authentication
 directory.

healthvault-ios-sdk/Classes/Controllers/MHVBrowserAuthBroker.m
```diff
@@ -20,13 +20,14 @@
...
-typedef void (^MHVSignInCompletion)(NSURL *_Nullable successUrl, NSError *_Nullable error);
-
 @interface MHVBrowserAuthBroker ()
 
 @property (nonatomic, strong) MHVSignInCompletion completion;
 @property (nonatomic, strong) dispatch_queue_t authQueue;
 @property (nonatomic, strong) NSURL *endUrl;
+@property (nonatomic, strong) NSURL *startUrl;
+
+@property (nonatomic, strong) MHVBrowserController *browserController;
```
namalu: We should probably move this out of the Controllers directory and put it in the Authentication
 directory.

healthvault-ios-sdk/Classes/Controllers/MHVBrowserAuthBroker.m
```diff
@@ -91,85 +94,84 @@ - (void)showAuthenticationControllerFromViewController:(UIViewController *)viewC
...
-- (void)cancelButtonPressed:(id)sender
+- (void)userCancelled
 {
-    [self completeWithSuccessUrl:nil error:[NSError error:[NSError MHVOperationCancelled] withDescription:@"The operation has been cancelled by the user."]];
+    [self completeWithSuccessUrl:nil
+                           error:[NSError error:[NSError MHVOperationCancelled] withDescription:@"The operation has been cancelled by the user."]];
 }
 
-#pragma mark - UIWebViewDelegate
+#pragma mark - WKWebViewDelegate
```
namalu: I think we need to add in didReceiveAuthenticationChallenge:, didFailProvisionalNavigation:, and webViewWebContentProcessDidTerminate:. I want to make sure that we don't orphan this completion block.

healthvault-ios-sdk/Classes/Controllers/MHVBrowserAuthBrokerProtocol.h
```diff
@@ -17,17 +17,35 @@
...
-
+#import <WebKit/WebKit.h>
 @class UIViewController;
 
+typedef void (^MHVSignInCompletion)(NSURL *_Nullable successUrl, NSError *_Nullable error);
+
 NS_ASSUME_NONNULL_BEGIN
 
-@protocol MHVBrowserAuthBrokerProtocol <NSObject>
+@protocol MHVBrowserAuthBrokerProtocol <NSObject, WKNavigationDelegate>
```
namalu: We should probably move this out of the Controllers directory and put it in the Authentication
directory.

healthvault-ios-sdk/Classes/Controllers/MHVBrowserController.m
```diff
@@ -18,166 +18,103 @@
...
-    frame.origin.x = 0;
-    frame.origin.y = 0;
-    
-    self.webView = [[UIWebView alloc] initWithFrame:frame];
-    
+    WKWebViewConfiguration *configuration = [[WKWebViewConfiguration alloc] init];
+    configuration.websiteDataStore = [WKWebsiteDataStore nonPersistentDataStore];
+    configuration.dataDetectorTypes = WKDataDetectorTypeNone;
+
+    self.webView = [[WKWebView alloc] initWithFrame:self.view.bounds configuration:configuration];
```
namalu: We should use auto layout here and set-up constraints.

healthvault-ios-sdk/Classes/Controllers/MHVBrowserController.m
```diff
@@ -186,12 +123,34 @@ - (void)cancelButtonPressed:(id)sender
...
-        [self.navigationController popViewControllerAnimated:TRUE];
+        [self.authBroker userCancelled];
     }
 }
 
+#pragma mark - Activity spinner
+
+- (void)showActivitySpinner
+{
+    if (!self.activityView)
```
namalu: Here too - Auto layout.

healthvault-ios-sdk/Classes/Controllers/MHVBrowserController.m
```diff
@@ -200,4 +159,47 @@ - (void)hideActivitySpinner
...
+#pragma mark - KVO on webView.loading to show spinner
+
+- (void)startObserving
+{
+    [self.webView addObserver:self forKeyPath:kLoadingKeyPath options:kNilOptions context:nil];
+}
+
+- (void)stopObserving
+{
+    [self.webView removeObserver:self forKeyPath:kLoadingKeyPath];
```
namalu: You need to check if self.webview is nil before removing the observer. This can cause a crash.

healthvault-ios-sdk/Classes/Controllers/MHVBrowserAuthBroker.m
```diff
@@ -91,85 +94,84 @@ - (void)showAuthenticationControllerFromViewController:(UIViewController *)viewC
             vc = [[[UIApplication sharedApplication] keyWindow] rootViewController];
         }
         
-        UINavigationController *navigationController = [[UINavigationController alloc] initWithRootViewController:self];
+        if (!self.browserController)
```
namalu: I actually think we should create a new browser controller every time. I'm noticing that the previous web view displays old content while navigating to the new page. I see that you are hiding the web view by setting the alpha to zero, but there might still be some leftover content from a previous session.

namalu (changes requested):

healthvault-ios-sdk/Classes/Controllers/MHVBrowserAuthBroker.m
miburf: It does do that.  It sets browserController to nil when it completes and calls the completion block.  It doesn't create if it is set so test cases can setup a mock browser controller.  I didn't want a test init that would be confusing for people using the SDK.  Added comments

healthvault-ios-sdk/Classes/Controllers/MHVBrowserController.m
miburf: self.webview nil should do nothing; though the observer:self being nil would cause problems

healthvault-ios-sdk/Classes/Controllers/MHVBrowserAuthBrokerProtocol.h
miburf: moved

namalu (approved):

PR #89, opened 2017-06-15
Implement methods in Person Client

miburf:
Implement methods in Person Client
Add tests for person client
Add person client features to SDK sample app


Example/Tests/Clients/MHVPersonClientTests.m
```diff
@@ -0,0 +1,476 @@
...
+    beforeEach(^{
+        returnedSettings = nil;
+        returnedPersonInfo = nil;
+        returnedRecords = nil;
+        returnedError = nil;
+    });
+    
+#pragma mark - GetApplicationSettings
+    
+    context(@"GetApplicationSettings success", ^
```
namalu: Same feedback here about BDD test strings. The test name should end up being an English-Like sentence.

healthvault-ios-sdk/Classes/Clients/MHVPersonClientProtocol.h
```diff
@@ -18,38 +18,72 @@
...
- * @param completion Envoked when the operation completes. NSArray<MHVPersonInfo *> an array of MHVPersonInfo objects representing people authorized for the application. NSError object will be nil if there is no error when performing the operation.
+ *        This allows continuing if a previous getAuthorizedPeople call had the moreResults flag set
+ * @param completion Envoked when the operation completes. 
+ *        MHVGetAuthorizedPeopleResult containing an array of MHVPersonInfo objects representing people authorized for the application, and a moreResults flag.
+ *        NSError object will be nil if there is no error when performing the operation.
  */
-- (void)getAuthorizedPeopleWithSettings:(MHVGetAuthorizedPeopleSettings *_Nonnull)settings
-                             completion:(void(^)(NSArray<MHVPersonInfo *> *_Nullable personInfos, NSError *_Nullable error))completion;
+- (void)getAuthorizedPeopleWithSettings:(MHVGetAuthorizedPeopleSettings *)settings
+                             completion:(void(^)(MHVGetAuthorizedPeopleResult *_Nullable authorizedPeople, NSError *_Nullable error))completion;
```
namalu: I don't think we should change the return type from and array of MHVPersonInfos to the MHVGetAuthorizedPeopleResult. We need to handle the case where there are "more results" internally, and not expose this internal object to the caller.

Example/Tests/Connection/MHVSodaConnectionTests.m
```diff
@@ -1098,7 +1100,7 @@
                 
                 it(@"should forward the error to the caller", ^
                    {
-                       [[expectFutureValue(expectedError) shouldEventually] beNonNil];
+                       [[expectFutureValue(expectedError) shouldEventuallyBeforeTimingOutAfter(3)] beNonNil];
```
namalu: What happened here? Why is it now taking 3 seconds to complete? I'm not excited about adding delays to our tests because they will take longer to complete.

healthvault-ios-sdk/Classes/Clients/MHVPersonClient.m
```diff
@@ -71,55 +131,215 @@ - (void)getAuthorizedPeopleWithCompletion:(void(^)(NSArray<MHVPersonInfo *> *_Nu
...
+#pragma mark - Helpers
+
+- (void)getAuthorizedPeopleWithParameters:(NSString *_Nonnull)parameters
+                               completion:(void(^)(MHVGetAuthorizedPeopleResult *_Nullable authorizedPeople, NSError *_Nullable error))completion
+{
     MHVMethod *method = [MHVMethod getAuthorizedPeople];
-    method.parameters = @"<info><parameters></parameters></info>";
+    method.parameters = parameters;
     
     [self.connection executeHttpServiceOperation:method
```
namalu: Unfortunately, we can't call this done until we handle the case where there the result is incomplete (moreResults is YES). You are going to have to track down how this actually works.

namalu (changes requested):

healthvault-ios-sdk/Classes/Clients/MHVPersonClientProtocol.h
miburf: changed, and handling more results internally

Example/Tests/Connection/MHVSodaConnectionTests.m
miburf: Not sure, I was having consistent errors with some of the tests, but seem to be OK now so removed

Example/Tests/Clients/MHVPersonClientTests.m
```diff
@@ -0,0 +1,535 @@
...
+         completion(serviceResponse, serviceResponse.error);
+         return nil;
+     }];
+    
+    let(personClient, ^
+        {
+            return [[MHVPersonClient alloc] initWithConnection:mockConnection];
+        });
+    
+    beforeEach(^{
```
namalu: formatting: New line for curly bracket.

HealthVault/Classes/Clients/MHVPersonClient.m
```diff
@@ -53,13 +56,70 @@ - (instancetype)initWithConnection:(id<MHVConnectionProtocol>)connection
...
+    {
+        if (completion)
+        {
+            completion([NSError MVHRequiredParameterIsNil]);
+        }
+        return;
+    }
     
+    MHVMethod *method = [MHVMethod setApplicationSettings];
+    method.parameters = [self bodyForSetApplicationSettings:applicationSettings];
```
namalu: Because this must be valid XML, we should check if bodyForSetApplicationSettings: returns nil, and if so, provide the caller with a detailed error explaining to them that the string is not valid XML.

HealthVault/Classes/Clients/MHVPersonClient.m
```diff
@@ -53,13 +56,70 @@ - (instancetype)initWithConnection:(id<MHVConnectionProtocol>)connection
...
+             return;
+         }
+         
+         completion(applicationSettings, nil);
+     }];
 }
 
-- (void)setApplicationSettingsWithRequestParameters:(NSString *)requestParameters
-                                         completion:(void(^_Nullable)(NSError *_Nullable error))completion
+- (void)setApplicationSettings:(MHVApplicationSettings *)applicationSettings
```
namalu: I think you should just leave this parameter as "requestParameters". The current code forces the caller to create a string and create an MHVApplicationSettings object - It's just more work for the consumer. It's fine to use the MHVApplicationSettings internally.

HealthVault/Classes/Clients/MHVPersonClient.m
```diff
@@ -70,56 +130,274 @@ - (void)getAuthorizedPeopleWithCompletion:(void(^)(NSArray<MHVPersonInfo *> *_Nu
...
+    [writer writeStartElement:@"info"];
+    
+    for (NSUUID *uuid in recordIds)
+    {
+        [XSerializer serialize:[[MHVString alloc] initWith:uuid.UUIDString] withRoot:@"id" toWriter:writer];
+    }
+
+    [writer writeEndElement];
     
+    return [writer newXmlString];
```
namalu, line 400: I think we need to return nil here (or some other error check) if the string is not valid XML. This code, allows the user to make the set settings call with invalid XML. This method writes an empty info object, and the call will complete successfully even though we didn't actually save anything.

HealthVault/Classes/Clients/MHVPersonClient.m
```diff
@@ -70,56 +130,274 @@ - (void)getAuthorizedPeopleWithCompletion:(void(^)(NSArray<MHVPersonInfo *> *_Nu
...
+    }
+
+    [self getAuthorizedPeopleWithSettings:settings
+                              personInfos:nil
+                               completion:completion];
+}
+
+- (void)getAuthorizedPeopleWithSettings:(MHVGetAuthorizedPeopleSettings *)settings
+                            personInfos:(NSArray<MHVPersonInfo *> *_Nullable)personInfos
+                             completion:(void(^)(NSArray<MHVPersonInfo *> *_Nullable personInfos, NSError *_Nullable error))completion
```
namalu: It looks like there is a MHVPersonInfoCollection - Maybe we should use that instead.

HealthVault/Classes/Clients/MHVPersonClientProtocol.h
```diff
@@ -18,38 +18,72 @@
...
  *
  * @param settings The MHVGetAuthorizedPeopleSettings object used to configure the results returned by this method.
- * @param completion Envoked when the operation completes. NSArray<MHVPersonInfo *> an array of MHVPersonInfo objects representing people authorized for the application. NSError object will be nil if there is no error when performing the operation.
+ *        This allows continuing if a previous getAuthorizedPeople call had the moreResults flag set
+ * @param completion Envoked when the operation completes. 
+ *        MHVGetAuthorizedPeopleResult containing an array of MHVPersonInfo objects representing people authorized for the application, and a moreResults flag.
+ *        NSError object will be nil if there is no error when performing the operation.
  */
-- (void)getAuthorizedPeopleWithSettings:(MHVGetAuthorizedPeopleSettings *_Nonnull)settings
+- (void)getAuthorizedPeopleWithSettings:(MHVGetAuthorizedPeopleSettings *)settings
```
namalu: I think we should break out the authorizationsCreatedSince property on MHVGetAuthorizedPeopleSettings and only use that argument in this method call. MHVGetAuthorizedPeopleSettings seems like an internal object to me because we are handling fetching more results if the result is incomplete. In other words, the caller has no use for batchSize or startingPersonId.

namalu (changes requested):

HealthVault/Classes/Clients/MHVPersonClient.m
miburf: I'll add a check to that any others, though since it uses the base class's validate, it won't be invalid for the application settings.

HealthVault/Classes/Clients/MHVPersonClient.m
miburf: requestParameters sounded weird, changed to just NSString applicationSettings

HealthVault/Classes/Clients/MHVPersonClientProtocol.h
miburf: Broke out so you just give the authorizationsCreatedSince date

HealthVault/Classes/Clients/MHVPersonClientProtocol.h
```diff
@@ -18,38 +18,71 @@
 
 #import "MHVClientProtocol.h"
 
-@class MHVApplicationSettings, MHVPersonInfo, MHVHealthRecordInfo, MHVGetAuthorizedPeopleSettings;
+@class MHVPersonInfo, MHVPersonInfoCollection, MHVGetAuthorizedPeopleResult, MHVRecord, MHVRecordCollection, MHVGetAuthorizedPeopleSettings;
```
namalu: These declarations need to be cleaned up. We are no longer using MHVGetAuthorizedPeopleResult or MHVGetAuthorizedPeopleSettings

HealthVault/Classes/Clients/MHVPersonClientProtocol.h
```diff
@@ -18,38 +18,71 @@
...
 
 /**
- * Gets information about people authorized for an application.
+ * Gets information about people authorized for an application whose authorization was created after a date.
  *
- * @param settings The MHVGetAuthorizedPeopleSettings object used to configure the results returned by this method.
- * @param completion Envoked when the operation completes. NSArray<MHVPersonInfo *> an array of MHVPersonInfo objects representing people authorized for the application. NSError object will be nil if there is no error when performing the operation.
+ * @param authorizationsCreatedSince Get authorized people whose authorization was created after date
+ * @param completion Envoked when the operation completes. 
+ *        MHVGetAuthorizedPeopleResult containing an array of MHVPersonInfo objects representing people authorized for the application, and a moreResults flag.
```
namalu: The documentation is out of date for the current interface.

HealthVault/Classes/Types/MHVPersonInfo.h
```diff
@@ -20,12 +20,13 @@
...
 #import "MHVType.h"
 #import "MHVRecord.h"
+#import "MHVApplicationSettings.h"
 
 @interface MHVPersonInfo : MHVType
 
 @property (readwrite, nonatomic, strong) NSUUID *ID;
 @property (readwrite, nonatomic, strong) NSString *name;
-@property (readwrite, nonatomic, strong) NSString *appSettingsXml;
+@property (readwrite, nonatomic, strong) MHVApplicationSettings *applicationSettings;
```
namalu: Should this be reverted back to a string? It looks like if we do MHVApplicationSettings can be an internal object.

namalu (approved):
There are just a few cleanup comments.

HealthVault/Classes/Types/MHVPersonInfo.h
miburf: Yeah, I can make it pass-through to MHVApplicationSettings  internally, to make it simpler for the SDK users

PR #98, opened 2017-06-20
Cleaned up podspec, readme, and old files

namalu:


README.md
```diff
@@ -6,92 +6,57 @@
 [![Platform](https://img.shields.io/cocoapods/p/healthvault-ios-sdk.svg?style=flat)](http://cocoapods.org/pods/healthvault-ios-sdk)
 
 # About
-**healthvault-ios-sdk** is an iOS library you can use to build applications that leverage the Microsoft HealthVault platform. **healthvault-ios-sdk** is actively used by the [Microsoft HealthVault for iPhone app](https://itunes.apple.com/us/app/microsoft-healthvault/id546835834?mt=8).
 
-**healthvault-ios-sdk** introduces a rich new HealthVault iOS client programming model. It includes built in serialization of most HealthVault data types, and built in support for HealthVault methods.
+**healthvault-ios-sdk** is an iOS framework you can use to build applications that leverage the Microsoft HealthVault platform. **healthvault-ios-sdk** is actively used by the [Microsoft HealthVault for iPhone app](https://itunes.apple.com/us/app/microsoft-healthvault/id546835834?mt=8).
```
miburf, line 10: Should we remove this?  After all the work, the HealthVault app isn't really using what's here now.

miburf (approved):

README.md
namalu: I think the idea is that a developer can download the HealthVault app and see the functionality HealthVault can offer. Yes, the SDK interfaces have changed significantly but the endpoints are still the same and the data types are the same. It still uses the same SDK, just a different version of it.

PR #100, opened 2017-06-23
Add EncryptedCoreData & SQLCipher; add GetRecordOperations method

miburf:
Add EncryptedCoreData & SQLCipher; 
Add GetRecordOperations method to test for offline caching


Example/healthvault-ios-sdk/MHVFeaturesConfiguration.m
```diff
@@ -24,7 +24,7 @@ + (MHVConfiguration *)configuration
 {
     MHVConfiguration *config = [MHVConfiguration new];
 
-    config.masterApplicationId = [[NSUUID alloc] initWithUUIDString:@"7c81b435-f091-4444-9534-59eb53da66c9"];
+    config.masterApplicationId = [[NSUUID alloc] initWithUUIDString:@"d6318dff-5352-4a10-a140-6c82c6536a3b"];
```
namalu: Let's make sure we don't merge this change in. We should make sure the SDK sample is using the correct app id.

HealthVault.podspec
```diff
@@ -26,4 +26,6 @@ The healthvault-ios-sdk framework simplifies developing apps that use the Micros
   s.xcconfig         = { 'HEADER_SEARCH_PATHS' => '$(inherited) $(SDKROOT)/usr/include/libxml2', 'OTHER_LDFLAGS' => '-lxml2' }
   s.frameworks       = 'UIKit', 'Security', 'MobileCoreServices', 'SystemConfiguration'
 
+  s.dependency 'EncryptedCoreData', '~> 3.1'
```
namalu: We should probably add this as a subspec, so consumers opt-out of the dependency. We have to think of a decent name for the subspec though... Maybe ThingCache or EncryptedThingCache? I think we should also create a default subspec that does not include the EncryptedCoreData framework, so developers who don't use it, have a lighter package.
https://guides.cocoapods.org/syntax/podspec.html#subspec

The HockeyApp SDK has a pretty good example of subspecs that remove dependencies, and set a default:
https://github.com/bitstadium/HockeySDK-iOS/blob/develop/HockeySDK.podspec

HealthVault/Classes/MHVThingCache.h
```diff
@@ -0,0 +1,13 @@
+//
+//  MHVThingCache.h
```
namalu: We should probably add this file and the .m to the 'Store' directory and not the root 'Classes' directory. Either that or create a new directory altogether (which might be a better idea). At some point we need to clean up our directory structure into public and private directories, so we don't expose internal functionalities of the framework to consumers.

HealthVault/Classes/Methods/MHVMethod.h
```diff
@@ -347,6 +347,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 + (MHVMethod *)updateExternalId;
 
+/**
+ Get record operations
+ 
+ @return An instance of MHVMethod.
+ */
++ (MHVMethod *)getRecordOperations;
```
namalu: Methods are actually listed alphabetically to make them easier to find. Let's keep this pattern going.

HealthVault/Classes/Methods/MHVMethod.m
```diff
@@ -266,4 +266,9 @@ + (MHVMethod *)updateExternalId
     return [[MHVMethod alloc] initWithName:@"UpdateExternalId" version:1 isAnonymous:NO];
 }
 
++ (MHVMethod *)getRecordOperations
```
namalu: Here too. Alphabetically.

namalu (changes requested):

namalu (approved):

PR #102, opened 2017-06-26
Make usable enums from the swagger docs

mikenevi:


Example/healthvault-ios-sdk/ViewControllers/ActionPlans/MHVActionPlansListViewController.m
```diff
@@ -116,16 +116,16 @@ - (IBAction)addActionPlan:(id)sender
...
     frequencyTask.trackingPolicy = policy;
-    frequencyTask.completionType = @"Frequency";
+    frequencyTask.completionType = MHVActionPlanTaskCompletionTypeEnum.Frequency;
     frequencyTask.frequencyTaskCompletionMetrics = metrics;
     
     MHVSchedule *schedule = [[MHVSchedule alloc] init];
-    schedule.reminderState = @"Off";
-    schedule.scheduledDays = @[@"Saturday", @"Sunday"];
+    schedule.reminderState = MHVScheduleReminderStateEnum.Off;
+    schedule.scheduledDays = @[MHVScheduleScheduledDaysEnum.Saturday, MHVScheduleScheduledDaysEnum.Sunday];
```
mikenevi: I especially like this one

HealthVault/Tools/codegen/template/model-header.mustache
```diff
@@ -18,12 +19,31 @@
...
+{{/vars}}
+{{/model}}
+
 @interface {{classname}} : MHVModelBase
 
 {{#vars}}
 {{#description}}/* {{{description}}} {{^required}}[optional]{{/required}}
  */{{/description}}
-@property(strong,nonatomic{{^required}},nullable{{/required}}) {{{ datatype }}} {{name}};
+@property(strong,nonatomic{{^required}},nullable{{/required}}) {{#isEnum}}{{#isContainer}}NSArray<{{/isContainer}}{{classname}}{{enumName}}{{#isContainer}}*>{{/isContainer}}*{{/isEnum}}{{^isEnum}}{{{ datatype }}}{{/isEnum}} {{name}};
```
mikenevi, line 46: starting to get unwieldy. Might need to split this into different cases if it gets any more complicated.

Example/healthvault-ios-sdk/ViewControllers/ActionPlanTasks/MHVActionPlanTaskListViewController.m
```diff
@@ -64,7 +64,7 @@ - (IBAction)addTask:(id)sender
 
     MHVActionPlanFrequencyTaskCompletionMetrics *metrics = [[MHVActionPlanFrequencyTaskCompletionMetrics alloc] init];
     metrics.occurrenceCount = @(1);
-    metrics.windowType = @"Daily";
+    metrics.windowType = MHVActionPlanFrequencyTaskCompletionMetricsWindowTypeEnum.Daily;
```
namalu: ...Invalidating caches and naming things; the two most difficult things in computer science.

HealthVault/Classes/Models/Generated/MHVActionPlan.h
```diff
@@ -29,13 +29,22 @@
...
+#import "MHVEnum.h"
 
 
 @protocol MHVActionPlan
 @end
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface MHVActionPlanCategoryEnum : MHVEnum
++(MHVActionPlanCategoryEnum *) Unknown;
```
namalu: 1. When using the dynamic enums, I believe that you don't need to specify unknown explicitly; The super class should add that for you.
2. Can we remove the space between the ending parenthesis and the name of the enum?
3. We should prefix the enum method name with MHV. We had an issue with the HVI app getting rejected because of collisions with other iOS framework properties.

HealthVault/Classes/Models/Generated/MHVActionPlan.m
```diff
@@ -88,5 +88,38 @@ + (NSDictionary *)objectParametersMap
...
+    return @{
+        @"Unknown": @"Unknown",
+        @"Health": @"Health",
+        @"Sleep": @"Sleep",
+        @"Activity": @"Activity",
+        @"Stress": @"Stress",
+    };
+}
 
++(MHVActionPlanCategoryEnum *) Unknown
```
namalu: I believe if you have an enum map, you don't need to do these initializer methods. You will have to silence the compiler warnings, but that will hopefully be pretty simple.

HealthVault/Tools/codegen/template/model-header.mustache
namalu: No kidding.

namalu (changes requested):

namalu (approved):

PR #103, opened 2017-06-28
Add Thing Cache

miburf:
Add ThingCache, and ThingCacheDatabase
Still need to return query results from the cache in GetThings


Example/healthvault-ios-sdk/MHVAppDelegate.m
```diff
@@ -35,6 +35,17 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
 }
 
+- (void)application:(UIApplication *)application performFetchWithCompletionHandler:(void (^)(UIBackgroundFetchResult result))completionHandler
+{
+    // If background fetch is enabled, tell the cache to sync the latest data from HealthVault
+    id<MHVSodaConnectionProtocol> connection = [[MHVConnectionFactory current] getOrCreateSodaConnectionWithConfiguration:[MHVFeaturesConfiguration configuration]];
+    
+    if (connection)
+    {
+        [connection.thingClient.cache syncWithCompletionHandler:completionHandler];
```
namalu: I think we need to think more generically. It would be a good idea to expose a method to perform background tasks on the connection, and possibly a bit mask or other parameter representing the various background operations to perform. This way if we add other background tasks we can just add them to the connection and not force the consumer to continue to add more nested sync calls here.

Example/healthvault-ios-sdk/ViewControllers/MHVTypeListViewController.m
```diff
@@ -251,6 +251,7 @@ - (void)setupList
     NSMutableArray<NSString *> *restTypeList = [[NSMutableArray alloc] init];
     
     //HealthVault Thing Types
+    //NOTE: if any new types are added here, they should also be added in MHVFeaturesConfiguration
```
namalu: Is there any way that the MHVFeaturesConfiguration and this class can just share the same Array?

Example/healthvault-ios-sdk/ViewControllers/MHVRecordListViewController.swift
```diff
@@ -138,6 +138,11 @@ class MHVRecordListViewController: UIViewController, UITableViewDelegate, UITabl
             return
         }
         
+        if let cache = self.connection.thingClient()?.cache
+        {
+            cache .startSyncing(forRecord: record.id, completion: nil)
```
namalu: I don't think we want the consumer to be troubled with starting/stoping the db sync process. We should probably just start syncing as soon as the login process is complete, behind the scenes.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+#import "MHVNetworkStatusProtocol.h"
+
+typedef void (^MHVBackgroundFetchResultCompletion)(UIBackgroundFetchResult result);
+
+@interface MHVThingCache ()
+
+@property (nonatomic, weak)   id<MHVConnectionProtocol>                             connection;
+@property (nonatomic, strong) MHVThingCacheConfiguration                            *cacheConfiguration;
+@property (nonatomic, strong) id<MHVThingCacheDatabaseProtocol>                     database;
+@property (nonatomic, strong) id<MHVNetworkStatusProtocol>                          networkStatus;
```
namalu: I think MHVNetworkObserverProtocol and networkObserver make more sense here. NetworkStatus seems like it would be an enum with the status of the network.

HealthVault/Classes/Caching/MHVThingCacheProtocol.h
```diff
@@ -0,0 +1,70 @@
...
+@protocol MHVConnectionProtocol, MHVNetworkStatusProtocol;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol MHVThingCacheProtocol <NSObject>
+
+/**
+ The application has been deauthorized.  All data for the caching should be deleted
+ */
+- (void)deauthorizedApplication;
```
namalu: This method does not deauthorize the application like the name implies. Maybe clearAllCachedDataWithCompletion:

We probably need a completion or some other way to communicate errors to the caller.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+#pragma mark - Cached Results
+
+- (void)cachedResultsForQueries:(MHVThingQueryCollection *)queries
+                       recordId:(NSUUID *)recordId
+                     completion:(void(^)(MHVThingQueryResultCollection *_Nullable resultCollection))completion;
+{
+    id<MHVCachedRecord> record = [self.database fetchCachedRecord:recordId.UUIDString];
+    if (!record)
+    {
+        completion(nil);
```
namalu: You need to check if completion is nil before calling it to prevent crashing.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+    {
+        [self.syncCompletionHandlers removeAllObjects];
+    }
+}
+
+#pragma mark - Cached Results
+
+- (void)cachedResultsForQueries:(MHVThingQueryCollection *)queries
+                       recordId:(NSUUID *)recordId
+                     completion:(void(^)(MHVThingQueryResultCollection *_Nullable resultCollection))completion;
```
namalu: We should add an error in the completion so we can provide errors to the caller if something goes wrong.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+{
+    id<MHVCachedRecord> record = [self.database fetchCachedRecord:recordId.UUIDString];
+    if (!record)
+    {
+        completion(nil);
+        return;
+    }
+
+    // If offline, dates don't apply & always use cached data
+    if (self.networkStatus.hasNetworkConnection)
```
namalu: I don't think we need to do this.The sync interval should be controlled by a periodic sync that is configurable by setting a property on the cache configuration. The app should sync if the period between the last sync >= now. Obviously if the app is backgrounded, this period might be longer, but we should respect the configuration.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+        }
+    }
+    
+    //TODO...use cache
+    completion(nil);
+}
+
+#pragma mark - Syncing
+
+- (void)startSyncingForRecordId:(NSUUID *)recordId completion:(void (^_Nullable)(BOOL success))completion
```
namalu: We need errors in the completion here. This is a public SDK and we need our error reporting to be exceptional. Just returning "NO" the sync didn't work will not help anyone debug there issues.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+        {
+            completion(NO);
+        }
+        return;
+    }
+    
+    id<MHVCachedRecord> record = [self.database fetchCachedRecord:recordId.UUIDString];
+    if (!record)
+    {
+        MHVLOG(@"Record not found, adding new Sync record %@", [recordId.UUIDString substringToIndex:4]);
```
namalu: Maybe we shouldn't log even a partial record id.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+    MHVASSERT_PARAMETER(recordId);
+
+    if (!recordId)
+    {
+        return;
+    }
+    [self.database deleteRecord:recordId.UUIDString];
+}
+
+- (void)syncWithCompletionHandler:(void (^)(UIBackgroundFetchResult result))completionHandler
```
namalu: Using UIBackgroundFetchResult is limiting your design here. I mentioned in the app delegate file, that there should be some abstraction between the cache and the performFetchWithCompletionHandler: method, and here is another example why. We need to provide more detail back to the caller in the form of errors or enums or something that will help debug issues, provide a way to test failures, and give the code a chance to react appropriately based on the problem. The fetchCachedRecords call to the database could have an error, but we are just logging it. That means that if there is an error we can't actually do anything to recover from it because we are just swallowing it.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+{
+    MHVASSERT_PARAMETER(completionHandler);
+
+    [self.database fetchCachedRecords:^(NSArray<id<MHVCachedRecord>> *_Nullable records)
+     {
+         if (records.count == 0)
+         {
+             MHVLOG(@"No Records to sync, done");
+             completionHandler(UIBackgroundFetchResultNoData);
```
namalu: You need to check if completion is nil before calling it to prevent crashes.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,410 @@
...
+         }
+         
+         [self syncRecords:records completion:completionHandler];
+     }];
+}
+
+- (void)syncRecords:(NSArray<id<MHVCachedRecord>> *_Nullable)records
+         completion:(void (^)(UIBackgroundFetchResult result))completion
+{
+    MHVASSERT_PARAMETER(completion);
```
namalu: Do we need to return here if completion is nil? or is completion optional?

HealthVault/Classes/Caching/MHVThingCacheConfiguration.h
```diff
@@ -0,0 +1,44 @@
...
+ To avoid downloading data that will never be used, this should be set with 
+ the data types used by the app.  If nil or empty, no data will be cached.
+ */
+@property (nonatomic, strong) NSArray<NSString *> *cacheTypeIds;
+
+/**
+ If the last time the cache was updated is older than this number, 
+ the cache is not used and the request is sent to HealthVault
+ */
+@property (nonatomic, assign) NSInteger maxCacheValidSeconds;
```
namalu: This property should be the periodic sync duration.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,448 @@
...
+        _lockObject = [NSObject new];
+    }
+    return self;
+}
+
+- (void)deleteDatabase
+{
+    @synchronized (self.lockObject)
+    {
+        [[NSFileManager defaultManager] removeItemAtURL:self.databaseUrl error:nil];
```
namalu: The NSFileManager needs to be injected so that we can test that removeItem is called when we expect it.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,448 @@
...
+        
+        _deleted = YES;
+    }
+}
+
+#pragma mark - Create Objects
+
+- (MHVCachedThing *)newThingForRecord:(MHVCachedRecord *)record
+{
+    if (self.deleted)
```
namalu: This isn't thread safe. It's possible that this method is called on a background thread, is paused after it makes it past the check of self.deleted, and the database then gets deleted.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,448 @@
...
+        return nil;
+    }
+    
+    __block NSError *error = nil;
+    
+    [self.managedObjectContext performBlockAndWait:^
+    {
+        if ([self.managedObjectContext hasChanges] && ![self.managedObjectContext save:&error])
+        {
+            MHVLOG(@"Error: %@", error.localizedDescription);
```
namalu: It isn't sufficient for us to just log the error, we need to return this error to the caller for debugging purposes and testing purposes.

namalu (changes requested):

HealthVault/Classes/Clients/MHVThingClient.m
```diff
@@ -146,8 +150,34 @@ - (void)getThingsWithQueries:(MHVThingQueryCollection *)queries
...
     }
-
+    
+#ifdef THING_CACHE
+    // Check for cached results for the GetThings queries
+    if (self.cache)
+    {
+        [self.cache cachedResultsForQueries:queries
+                                   recordId:recordId
+                                 completion:^(MHVThingQueryResultCollection * _Nullable resultCollection)
```
namalu: Here is a perfect example of why we need errors reported back to the caller. If an error occurs while trying to fetch the cached records, we need could try and fetch from the cloud. Instead, we don't know if there is an error, and we just always make a cloud request if there are no results.

HealthVault/Classes/Clients/MHVThingClient.m
```diff
@@ -146,8 +150,34 @@ - (void)getThingsWithQueries:(MHVThingQueryCollection *)queries
...
+         {
+             if (resultCollection)
```
namalu: We should just return the empty collection to the caller. Our model is eventual consistency, and we should let the periodic sync do it's job. We can fetch from the cloud if an error occurs with the cache, but it doesn't seem appropriate to just to make the request because we didn't find anything. This wouldn't handle cases where a caller requested all blood pressure things and we had cached 1 blood pressure, but 3 more had been added since the last sync. If anything, we might want to add a property to the MHVThingQuery that enables callers to bypass the cache so they could add a refresh button or pull to refresh to force a specific query.

HealthVault/Classes/Configuration/MHVConfiguration.h
```diff
@@ -108,4 +109,9 @@
  */
 @property (nonatomic, assign) BOOL isMultiRecordApp;
 
+/**
+ Configuration for caching
+ */
+@property (nonatomic, strong) MHVThingCacheConfiguration *cacheConfiguration;
```
namalu: This property should be on the MHVConnection. It needs to be separate from MHVConfiguration. The reason is that the MHVConfiguration data will, at some point, be loaded from an auto generated .json file that consumers get from the dashboard of their HealthVault account. The cache configuration has nothing to do with the app configuration.

HealthVault/Classes/Connection/MHVConnectionProtocol.h
```diff
@@ -41,6 +41,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, readonly, nullable) MHVSessionCredential *sessionCredential;
 
 /**
+ The configuration object for the current session.
+ */
+@property (nonatomic, strong, readonly) MHVConfiguration *configuration;
```
namalu: We should move this back to the MHVConnection and separate the cache configuration and the app configuration.

HealthVault/Classes/HTTP/MHVNetworkReachability.m
```diff
@@ -137,78 +117,78 @@ -(BOOL)startMonitoring
...
+        shouldNotify = (flags != 0 && self.status != flags);
+        self.status = flags;
     }
     if (shouldNotify)
     {
-        [[NSNotificationCenter defaultCenter]
-         postNotificationName: MHVHostReachabilityNotificationName
-         object:self
-         ];
+        [[NSNotificationCenter defaultCenter] postNotificationName:MHVHostReachabilityNotificationName
```
namalu: Let's get rid of the notifications before someone starts using them. We should just inject the network observer into the classes that need it. Notifications cause problems with tests, and can open the door for less than ideal architecture.

HealthVault/Classes/HTTP/MHVNetworkReachability.m
```diff
@@ -58,76 +38,76 @@ static void HostReachabilityStatusChanged(SCNetworkReachabilityRef target, SCNet
...
     }
     
     return TRUE;
 }
 
--(BOOL)startMonitoring
+- (BOOL)startMonitoring
 {
-    if (m_isMonitoring)
+    if (self.isMonitoring)
```
namalu: This isn't thread safe. We should fix it if we plan on using this class. The setting and checking of the isMonitoring BOOL needs to be synchronized.

HealthVault/Classes/Store/MHVKeychainServiceProtocol.h
```diff
@@ -66,6 +66,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (BOOL)removeObjectForKey:(NSString *)key;
 
+/**
+ Generate a very long random password, suitable for using as Cache database encryption key
+ 
+ @return The random string.
+ */
+- (NSString *)generateRandomPassword;
```
namalu: This doesn't belong in the keychain service. You should create a new class that does this or just add it as a method to the MHVThingCacheDatabase.

HealthVault/Classes/Types/MHVThingQuery.h
```diff
@@ -44,6 +44,7 @@
 
 @property (readwrite, nonatomic) int maxResults;
 @property (readwrite, nonatomic) int maxFullResults;
+@property (readwrite, nonatomic) BOOL ignoreCache;
```
namalu: maybe shouldUseCachedResults would be more clear. Also, we should explicitly set a default value when this object gets initialized.

namalu (changes requested):

HealthVault/Classes/Caching/MHVThingCache.h
```diff
@@ -1,13 +1,37 @@
 //
 //  MHVThingCache.h
```
namalu: This class should be in the 'private' directory.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,471 @@
...
+        {
+            _syncTypes = [[NSSet alloc] initWithArray:_cacheConfiguration.cacheTypeIds];
+        }
+    }
+    return self;
+}
+
+- (void)startCache
+{
+    self.syncTimer = [NSTimer scheduledTimerWithTimeInterval:_cacheConfiguration.syncIntervalSeconds
```
namalu: Rather than timer that repeats, maybe we should save the last successful sync completion time and sync after that. I'm thinking if a sync takes a long time to complete, we could end up syncing again before the first one is done.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,471 @@
...
+        
+        if (_cacheConfiguration.cacheTypeIds.count > 0)
+        {
+            _syncTypes = [[NSSet alloc] initWithArray:_cacheConfiguration.cacheTypeIds];
+        }
+    }
+    return self;
+}
+
+- (void)startCache
```
namalu: I think it would be cleaner to use KVO to monitor changes to the personInfo property on the connection. Rather than starting and stopping the sync process with method calls, if we just watch for changes to the personInfo.records array, we will know exactly when to start syncing and when to delete records that have been deauthorized.

HealthVault/Classes/Caching/MHVThingCacheDatabase.h
```diff
@@ -0,0 +1,29 @@
+//
+//  MHVThingCacheDatabase.h
```
namalu: Should be in the 'Private' directory.

HealthVault/Classes/Clients/MHVThingClientProtocol.h
```diff
@@ -20,12 +20,14 @@
...
 
 @class MHVThing, MHVThingQuery, MHVThingCollection, MHVThingQueryCollection, MHVThingQueryResultCollection, MHVBlobPayloadThing, MHVGetRecordOperationsResult;
-@protocol MHVBlobSourceProtocol;
+@protocol MHVThingCacheProtocol, MHVBlobSourceProtocol;
 
 NS_ASSUME_NONNULL_BEGIN
 
 @protocol MHVThingClientProtocol <NSObject>
 
+@property (readonly, nonatomic, strong) id<MHVThingCacheProtocol>   cache;
```
namalu: I don't think cache property should be in the publicly exposed protocol. We don't want to expose the internal cache to consumers. Since this is something that they should not have access to, It needs to be made private by either referencing the implementation of the thing client (MHVThingClient) in the connection and add the property there.

HealthVault/Classes/Connection/MHVConnectionFactoryInternal.m
```diff
@@ -62,6 +62,7 @@ - (instancetype)init
         if (!self.connection)
         {
             self.connection = [[MHVSodaConnection alloc] initWithConfiguration:configuration
+                                                            cacheConfiguration:nil
```
namalu: We should initialize this with a default cache configuration and then allow consumers to update properties on the object. We are adding an additional step for developers that want to enable thing caching when we should just default and let them override the default as they see fit.

HealthVault/Classes/Connection/MHVConnectionProtocol.h
```diff
@@ -100,6 +118,25 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ Start all the Cache processes.  Should be called after user is authenticated
+ */
+- (void)startCaches;
+
+/**
+ Remove all Cached data.  Should be called after a user is deauthorized
+ */
+- (void)clearAllCachedData;
```
namalu: I don't think startCaches and clearAllCachedData should be exposed publicly. These are operations we should handle internally. (See my comments about using KVO on the personInfo.records).

HealthVault/Classes/Connection/MHVConnectionProtocol.h
```diff
@@ -100,6 +118,25 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (id<MHVVocabularyClientProtocol> _Nullable)vocabularyClient;
 
+/**
+ Sync cached data
+
+ @param dataTypes Data types to be synced
+ @param options Sync options
+ @param completion Envoked when the sync process is complete, with count of items synced and any errors that occurred
+ */
+- (void)syncDataTypes:(MHVSyncDataTypes)dataTypes options:(MHVSyncOptions)options completion:(void(^_Nullable)(NSInteger syncedItemCount, NSError *_Nullable error))completion;
```
namalu: We should consider naming this performBackgroundTasks - I want it to be obvious that this method should be added to the performFetchWithCompletionHandler: and that this is specifically for background execution... and we should be as generic as possible. There may be other tasks we want to run in the background that have nothing to do with syncing or items, or types, or even data. I think the should return an object that summarizes what operations were performed and if it was a success.

HealthVault/Classes/Connection/MHVConnectionProtocol.h
```diff
@@ -19,10 +19,22 @@
 #import <Foundation/Foundation.h>
 #import <UIKit/UIKit.h>
 
-@class MHVSessionCredential, MHVPersonInfo, MHVServiceResponse, MHVMethod, MHVRemoteMonitoringClient;
+@class MHVSessionCredential, MHVConfiguration, MHVThingCacheConfiguration, MHVPersonInfo, MHVServiceResponse, MHVMethod, MHVRemoteMonitoringClient;
 
 @protocol MHVHttpServiceOperationProtocol, MHVPersonClientProtocol, MHVPlatformClientProtocol, MHVThingClientProtocol, MHVVocabularyClientProtocol, MHVRemoteMonitoringClient;
 
+typedef NS_ENUM(NSInteger, MHVSyncDataTypes)
```
namalu: MHVBackgroundTaskType - This may represent tasks that do not involve syncing.

HealthVault/Classes/Connection/MHVConnectionProtocol.h
```diff
@@ -19,10 +19,22 @@
...
+{
+    MHVSyncDataTypesThings = 0x1,
+    MHVSyncDataTypesAll = (MHVSyncDataTypesThings),
+};
+
+typedef NS_ENUM(NSInteger, MHVSyncOptions)
```
namalu: Not sure this is needed if we name the method that uses it 'background'.

HealthVault/Classes/Configuration/MHVConfigurationConstants.h
```diff
@@ -64,4 +64,9 @@ static NSTimeInterval const kDefaultRequestTimeoutDurationInSeconds = 30;
  */
 static NSInteger const kDefaultBlobChunkSizeInBytes = 1 << 21; // 2Mb.
 
+/*
+ The sync time in seconds
+ */
+static NSInteger const kDefaultSyncIntervalSeconds = 60*60; // 1 hour
```
namalu: I think this belongs in the cache configuration object.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,471 @@
...
+#pragma mark - Cached Results
+
+- (void)cachedResultsForQueries:(MHVThingQueryCollection *)queries
+                       recordId:(NSUUID *)recordId
+                     completion:(void(^)(MHVThingQueryResultCollection *_Nullable resultCollection, NSError *_Nullable error))completion;
+{
+    id<MHVCachedRecord> record = [self.database fetchCachedRecord:recordId.UUIDString];
+    if (!record)
+    {
+        completion(nil, [NSError error:[NSError MHVNotFound] withDescription:@"recordId not found in cache database"]);
```
namalu: We need to check if the completion is nil before calling it to avoid crashes.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,471 @@
...
+                                   {
+                                       completion([NSError error:[NSError MHVOperationCannotBePerformed]
+                                                 withDescription:@"Record could not be created in Cache database"]);
+                                   }
+                                   return;
+                               }
+                           }
+                       }
+
+                       completion(nil);
```
namalu: Here too. Check if completion is nil.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+- (NSError *_Nullable)deleteDatabase
+{
+    NSError *error = nil;
+    
+    @synchronized (self.lockObject)
+    {
+        [self.fileManager removeItemAtURL:self.databaseUrl
+                                    error:&error];
+        
+        [self.keychainService setString:nil forKey:kMHVCachePasswordKey];
```
namalu: nit: [self.keychainService removeObjectForKey: kMHVCachePasswordKey]

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+    }
+    
+    return error;
+}
+
+#pragma mark - Create Objects
+
+- (MHVCachedThing *)newThingForRecord:(MHVCachedRecord *)record
+{
+    if (self.deleted || !record)
```
namalu: Need to be synchronized - Not thread safe.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+                                               inManagedObjectContext:self.managedObjectContext];
+         thing.record = record;
+     }];
+    
+    return thing;
+}
+
+- (MHVCachedRecord *)newRecord:(NSString *)recordId
+{
+    if (self.deleted || !recordId)
```
namalu: Need to be synchronized - Not thread safe.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+    return record;
+}
+
+#pragma mark - Delete
+
+- (void)deleteRecord:(NSString *)recordId
+{
+    MHVASSERT_PARAMETER(recordId);
+    
+    if (self.deleted || !recordId)
```
namalu: Need to be synchronized - Not thread safe.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+         }
+     }];
+}
+
+- (NSError *_Nullable)deleteThingIds:(NSArray<NSString *> *)thingIds recordId:(NSString *)recordId
+{
+    MHVASSERT_PARAMETER(thingIds);
+    MHVASSERT_PARAMETER(recordId);
+    
+    if (self.deleted || !thingIds || thingIds.count == 0)
```
namalu: Need to be synchronized - Not thread safe. Also, it would appear that the thingIds parameter is required (and required that there be at least one element) - we should return an error here.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+}
+
+- (void)addOrUpdateThings:(MHVThingCollection *)things
+                 recordId:(NSString *)recordId
+       lastSequenceNumber:(NSInteger)lastSequenceNumber
+               completion:(void (^)(NSInteger updateItemCount, NSError *_Nullable error))completion
+{
+    MHVCachedRecord *record = (MHVCachedRecord *)[self fetchCachedRecord:recordId];
+    
+    if (self.deleted)
```
namalu: Need to be synchronized - Not thread safe.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+             completion(results);
+         }
+     }];
+}
+
+- (MHVCachedRecord *)fetchCachedRecord:(NSString *)recordId
+{
+    __block MHVCachedRecord *record;
+    
+    if (self.deleted)
```
namalu: Need to be synchronized - Not thread safe.

HealthVault/Classes/Caching/MHVThingCache.m
```diff
@@ -1,20 +1,471 @@
...
+    return error;
+}
+
+#pragma mark - Cached Results
+
+- (void)cachedResultsForQueries:(MHVThingQueryCollection *)queries
+                       recordId:(NSUUID *)recordId
+                     completion:(void(^)(MHVThingQueryResultCollection *_Nullable resultCollection, NSError *_Nullable error))completion;
+{
+    id<MHVCachedRecord> record = [self.database fetchCachedRecord:recordId.UUIDString];
```
namalu: We need to add a check here to see if the database is ready to use. It should probably return an error that the thing client can use to issue the request to HealthVault until the database has been synchronized.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+- (NSURL *)databaseUrl
+{
+    @synchronized(self.lockObject)
+    {
+        if (self.deleted)
+        {
+            return nil;
+        }
+        
+        if (!_databaseUrl)
```
namalu: It would simplify the synchronization code significantly if, instead of lazy initialization of databaseUrl, persistentStoreCoordinator, objectModel, and managedObjectContext you just set them in a single method and called it when the object is initialized, and after delete.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+       lastSequenceNumber:(NSInteger)lastSequenceNumber
+               completion:(void (^)(NSInteger updateItemCount, NSError *_Nullable error))completion
+{
+    MHVCachedRecord *record = (MHVCachedRecord *)[self fetchCachedRecord:recordId];
+    
+    if (self.deleted)
+    {
+        if (completion)
+        {
+            completion(0, nil);
```
namalu: It seems like this is an error. We are asking to add or update things, but the database has been deleted?? I think we need to make an error and provide it to the completion block.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+        }
+        return;
+    }
+    
+    if (!record)
```
namalu: Is this supposed to be 'recordId'?

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+    MHVASSERT_PARAMETER(recordId);
+    
+    if (self.deleted || !thingIds || thingIds.count == 0)
+    {
+        return nil;
+    }
+    
+    if (!recordId)
+    {
+        return [NSError MVHRequiredParameterIsNil];
```
namalu: There are 2 parameters here and 2 error cases (thingIds is nil or empty, recordId is nil or empty). It would be helpful if the error has a detailed message so the caller would not have to guess what they did wrong.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+{
+    MHVASSERT_PARAMETER(thingIds);
+    MHVASSERT_PARAMETER(recordId);
+    
+    if (self.deleted || !thingIds || thingIds.count == 0)
+    {
+        return nil;
+    }
+    
+    if (!recordId)
```
namalu: isNilOrEmpty is the condition we should check for.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+            completion(0, [NSError MVHRequiredParameterIsNil]);
+        }
+        return;
+    }
+    
+    [self.managedObjectContext performBlock:^
+     {
+         for (MHVThing *thing in things)
+         {
+             MHVCachedThing *cachedThing = [record findThingWithThingId:thing.thingID];
```
namalu: What happens here if record is nil? I see you have a check for it above, but it returns a MVHRequiredParameterIsNil error which I assume is for the recordId parameter.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+
+- (void)fetchCachedRecords:(void(^)(NSArray<id<MHVCachedRecord>> *_Nullable records))completion
+{
+    MHVASSERT_PARAMETER(completion);
+    
+    if (self.deleted)
+    {
+        if (completion)
+        {
+            completion(nil);
```
namalu: This this condition too seems to warrant returning an error to the caller.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+        }
+        return;
+    }
+    
+    [self.managedObjectContext performBlock:^
+     {
+         NSFetchRequest *request = [NSFetchRequest fetchRequestWithEntityName:@"MHVCachedRecord"];
+         
+         NSError *error = nil;
+         NSArray *results = [self.managedObjectContext executeFetchRequest:request error:&error];
```
namalu: There is an error here too that should be returned to the caller.

HealthVault/Classes/Caching/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,570 @@
...
+        return nil;
+    }
+    
+    [self.managedObjectContext performBlockAndWait:^
+     {
+         NSFetchRequest *request = [NSFetchRequest fetchRequestWithEntityName:@"MHVCachedRecord"];
+         request.predicate = [NSPredicate predicateWithFormat:@"recordId == %@", recordId];
+         
+         NSError *error = nil;
+         NSArray *results = [self.managedObjectContext executeFetchRequest:request error:&error];
```
namalu: Here too. We need to report these errors back to the caller for traceability and program flow. The caller of this method has no idea if there was an error, or if the is just no record. prepareCacheForRecords: calls this and creates a new record if it is nil - That's not the appropriate behavior if there is an error fetching.

HealthVault/Classes/Caching/MHVThingCacheDatabaseProtocol.h
```diff
@@ -0,0 +1,156 @@
+//
+//  MHVThingCacheDatabaseProtocol.h
```
namalu, line 2: This looks like it belongs in the private directory. I think the MHVThingCacheProtocol should be exposed to the consumer in the cache configuration, and they can override the default behavior from there. Our implementation of encrypted core data is too specific to our implementation and provides no flexibility for developers to use SQL if they choose.

HealthVault/Classes/Caching/MHVThingCacheDatabaseProtocol.h
```diff
@@ -0,0 +1,156 @@
...
+
+@protocol MHVThingCacheDatabaseProtocol <NSObject>
+
+/**
+ Create a new cache record given a recordId
+
+ @param recordId The record ID
+ @return the cache record
+ */
+- (id<MHVCachedRecord>)newRecord:(NSString *)recordId;
```
namalu: We spoke offline about changing this to just using the recordId to provide more flexibility for consumers.

namalu (changes requested):

HealthVault/Classes/Caching/MHVThingCacheDatabaseProtocol.h
miburf: Moved so only protocols and configs are public

Example/Tests/Caching/MHVCacheQueryTests.m
```diff
@@ -497,6 +498,22 @@
...
+                           {
+                               MHVThingQuery *thingQuery = [[MHVThingQuery alloc] initWithThingID:@"TEST_THING_ID"];
+                               thingQuery.shouldUseCachedResults = NO;
+                               
+                               cacheQuery = [[MHVCacheQuery alloc] initWithQuery:thingQuery];
+                           });
+                
+                it(@"should set the canQueryCache property to NO", ^
+                   {
+                       [[theValue(cacheQuery.canQueryCache) should] beFalse];
```
namalu, line 514: We should verify the other properties: predicate is nil, error is nil.

Example/healthvault-ios-sdk/MHVAppDelegate.m
```diff
@@ -19,6 +19,8 @@
 
 #import "MHVAppDelegate.h"
 #import "MHVViewController.h"
+#import "MHVConnectionProtocol.h"
+#import "MHVSodaConnectionProtocol.h"
```
namalu, line 23: nit: These imports shouldn't be needed. They are imported through the healthvault-ios-sdk-Prefix.pch file.

HealthVault/Classes/Caching/MHVThingCacheDatabaseProtocol.h
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ Update a record with a new date and/or sequence number
+
+ @param record The record
+ @param lastSyncDate NSDate to update, should not update date on the record if nil
+ @param sequenceNumber NSNumber sequence number, should not update number on the record if nil
+ @return NSError if any error occurred
+ */
+- (NSError *_Nullable)updateRecordId:(NSString *)record lastSyncDate:(NSDate *_Nullable)lastSyncDate sequenceNumber:(NSNumber *_Nullable)sequenceNumber;
```
namalu, line 139: nit 'recordId'

HealthVault/Classes/Caching/MHVThingCacheDatabaseProtocol.h
```diff
@@ -0,0 +1,143 @@
...
+@protocol MHVThingCacheDatabaseProtocol <NSObject>
+
+/**
+ Initialize and setup database if needed.
+ This is called after a user has authenticated, and may be called after deleteDatabase
+ when signing out and signing in again
+
+ @return Error if there was an error
+ */
+- (NSError *_Nullable)setupDatabase;
```
namalu, line 35: I know this might end up being a lot of work, but I think the method signatures for any method in this protocol should have a completion property. Right now everything works for core data, but we don't know if someone else's implementation will need to call asynchronous methods internally. The setupDatabase call is a perfect example, If my setup might take a few seconds, we are giving consumer's no choice but to block this thread until the setup is done.

HealthVault/Classes/Caching/Private/MHVCachedThing+Cache.m
```diff
@@ -0,0 +1,47 @@
...
+    self.version = thing.key.version;
+    self.thingType = thing.type.typeID;
+    
+    self.createDate = thing.created.when;
+    self.createdByAppId = thing.created.appID;
+    //self.createdByPersonId;
+    self.effectiveDate = thing.effectiveDate;
+    self.updateDate = thing.updated.when;
+    self.updatedByAppId = thing.updated.appID;
+    //self.updatedByPersonId;
```
namalu, line 37: Why are these commented out? Can you add some notes for context?

HealthVault/Classes/Caching/MHVThingCacheProtocol.h
```diff
@@ -0,0 +1,63 @@
...
+/**
+ Start the cache processes.  This should be called after the user has authenticated
+ */
+- (void)startCache;
+
+/**
+ Remove all cached data, for example if the app has been deauthorized
+ @return NSError if any error occurred
+ */
+- (NSError *_Nullable)clearAllCachedData;
```
namalu, line 38: It looks like startCache and clearAllCachedData do not need to be declared in this protocol any longer.

HealthVault/Classes/Caching/MHVThingCacheDatabaseProtocol.h
```diff
@@ -0,0 +1,143 @@
...
+- (NSError *_Nullable)deleteDatabase;
+
+/**
+ Determine if a record exists
+ 
+ @param recordId the RecordID to find
+ @return CachedRecord matching the ID or nil
+ @note This operation is synchronous
+ */
+- (BOOL)hasRecordId:(NSString *)recordId;
```
namalu, line 51: maybe hasDataForRecordId:?

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,703 @@
...
+    }
+    
+    return error;
+}
+
+#pragma mark - Create Objects
+
+- (MHVCachedThing *)newThingForRecord:(MHVCachedRecord *)record
+{
+    @synchronized (self.lockObject)
```
namalu, line 91: I don't think you need to synchronize the whole code block, just the part where you  are checking the value of self.deleted.

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,703 @@
...
+         }];
+    }
+}
+
+- (MHVCachedRecord *)fetchCachedRecord:(NSString *)recordId
+{
+    // Callers are using @synchronized (self.lockObject)
+    __block MHVCachedRecord *record;
+    
+    if (self.deleted)
```
namalu, line 392: You need to wrap just the if (self.deleted){} in @synchronized to ensure the correct value of the BOOL.

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -0,0 +1,703 @@
...
+        
+        return error;
+    }
+}
+
+#pragma mark - Core Data
+
+- (NSError *)saveContext
+{
+    // Callers are using @synchronized (self.lockObject)
```
namalu, line 579: There is too much code within the synchronized context. You only need/should synchronize the code you need, which in this case the checking the value of self.deleted.

HealthVault/Classes/Clients/MHVClientFactory.m
```diff
@@ -24,6 +24,11 @@
 #import "MHVSessionCredentialClient.h"
 #import "MHVVocabularyClient.h"
 #import "MHVCryptographer.h"
+#import "MHVThingCache.h"
```
namalu, line 27: Anything cache related will need to be wrapped in #ifdef THING_CACHE. If the cache subspec is not included these files will not exist.

namalu (approved):

HealthVault/Classes/Caching/MHVThingCacheDatabaseProtocol.h
miburf: This is bigger, I'll do on a separate PR with using the cache (I'll do all the changes from these comments in a separate PR so they're seen and verified.)

HealthVault/Classes/Caching/Private/MHVCachedThing+Cache.m
miburf: Added in the next PR

HealthVault/Classes/Caching/MHVThingCacheProtocol.h
miburf: removed

PR #105, opened 2017-06-29
Add timeline APIs and Models

mikenevi:


Example/healthvault-ios-sdk/ViewControllers/Timeline/MHVTimelineSnapshotViewController.m
```diff
@@ -0,0 +1,89 @@
...
+@property (strong, nonatomic) IBOutlet UITextField *numberOfTasksLabel;
+@property (strong, nonatomic) IBOutlet MHVStatusLabel *statusLabel;
+
+@end
+
+@implementation MHVTimelineSnapshotViewController
+
+- (id)initWithTypeClass:(Class)typeClass useMetric:(BOOL)metric
+{
+    self = [super init];
```
namalu, line 41: nit: We usually check if self is nil before setting properties on it. There is a rare possibility that the [super init] would return nil.

if (self)
{
    MHVConfiguration *config = MHVFeaturesConfiguration.configuration;
    self.connection = [[MHVConnectionFactory current] getOrCreateSodaConnectionWithConfiguration:config];
}

HealthVault/Classes/Models/MHVDateTimeBase.h
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import "MHVModelBase.h"
+
+@interface MHVDateTimeBase : MHVModelBase
+
+@property (nonatomic, strong) NSDate* date;
+
+- (NSString*)getDateFormatString;
```
namalu: In objective-c we don't use a lot of 'get' prefixes. There are a few methods where we use this to have consistent naming with the OneSDK, if this is one of them then we should leave it. Otherwise, 'dateFormatString' would be more in line with the language.

HealthVault/Classes/Models/MHVDateTimeBase.m
```diff
@@ -0,0 +1,51 @@
...
+- (instancetype)initWithObject:(id)object objectParameters:(NSObject *)parameters
+{
+    NSDateFormatter *dateFormatter = [NSDateFormatter new];
+    [dateFormatter setDateFormat:[self getDateFormatString]];
+    [dateFormatter setTimeZone:[NSTimeZone timeZoneForSecondsFromGMT:0]];
+    
+    self = [super init];
+    if (self)
+    {
+        self.date = [dateFormatter dateFromString:object];
```
namalu: instead of id should we type the parameter? It appears that object is a string we should type it appropriately. If this is some limitation with the auto-generated code, we should do a check to make sure the object is the correct class before passing it off as a string.

namalu (approved):

HealthVault/Classes/Models/MHVDateTimeBase.h
mikenevi: right, fixing.

HealthVault/Classes/Models/MHVDateTimeBase.m
mikenevi: Not a limitation of the auto-gen code. The function is inherited from MHVModelBase - I should actually remove the declaration from the .h file. Since it's overriding, I probably shouldn't change the parameter type. I'll add a check for the class.

PR #106, opened 2017-07-02
Working Caching

miburf:
Integrate Cache with ThingClient to return cached results
Add completions to database methods, and other PR comments
Add to Sample to see load time and reload with cache vs cloud


HealthVault/Classes/Caching/Private/MHVThingCache.m
```diff
@@ -104,27 +104,32 @@ - (void)startCache
...
+    [self.database deleteDatabaseWithCompletion:^(NSError * _Nullable error)
+     {
+         if (error)
+         {
+             MHVLOG(@"ThingCache: Error deleting database: %@", error.localizedDescription);
+         }
+         
+         @synchronized (self.syncCompletionHandlers)
+         {
+             [self.syncCompletionHandlers removeAllObjects];
```
namalu: Don't we need to call the completion on any pending completion handlers so call don't get orphaned?

HealthVault/Classes/Caching/Private/MHVThingCache.m
```diff
@@ -133,40 +138,159 @@ - (void)cachedResultsForQueries:(MHVThingQueryCollection *)queries
...
     {
-        if (completion)
-        {
-            completion(nil, [NSError error:[NSError MHVCacheNotReady] withDescription:@"recordId not found in cache database"]);
-        }
         return;
     }
     
-    if (![self.database isCacheValidForRecordId:recordId.UUIDString])
+    [self.database syncDataForRecordId:recordId.UUIDString
```
namalu: Maybe this should be named cacheStatusForRecordId:. I read this and think that this method triggers a sync.

HealthVault/Classes/Caching/Private/MHVThingCache.m
```diff
@@ -133,40 +138,159 @@ - (void)cachedResultsForQueries:(MHVThingQueryCollection *)queries
...
+{
+    for (MHVThing *thing in things)
     {
-        if (completion)
+        NSDate *date = [NSDate date];
+        if (created)
         {
-            completion(nil, [NSError MHVCacheError:@"Cache is not valid for record"]);
+            thing.created = [MHVAudit new];
+            thing.created.when = date;
```
namalu: We should probably check to make sure date is nil before setting it. In the HVI app, I know we can create things (weight & blood pressure) and set when to a date in the past.

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -63,7 +64,7 @@ - (instancetype)initWithKeychainService:(id<MHVKeychainServiceProtocol>)keychain
     return self;
 }
 
-- (NSError *_Nullable)deleteDatabase
+- (void)deleteDatabaseWithCompletion:(void (^)(NSError *_Nullable error))completion;
```
namalu: I think that this method should be 'resetDatabaseWithCompletion:' and it should delete the current database and the code in 'setupDatabaseWithCompletion:' should be merged into this method. I'm still seeing potential threading issues with the code in it's current state, that we can resolve by simply making sure there is always a database. You can still use self.lockObject to synchronize, but instead of a self.deleted BOOL checking for the existence of a database, you have a synchronized method to check the status of the database. All code in the 'resetDataBaseWithCompletion:' is wrapped in @synchronized(self.lockObject). Now you can create a single method to check the database status:

 - (BOOL)isDatabaseReady
{
    @synchronized (self.lockObject)
    { 
          return self.persistentStoreCoordinator != nil && self.managedObjectContext != nil; 
    }
}

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -94,145 +99,295 @@ - (MHVCachedThing *)newThingForRecord:(MHVCachedRecord *)record
...
+
+    if (!recordId)
+    {
+        completion(nil, [NSError MVHRequiredParameterIsNil]);
+        return;
+    }
+    
+    MHVCacheQuery *cacheQuery = [[MHVCacheQuery alloc] initWithQuery:query];
+    
+    if (cacheQuery.error)
```
namalu: This code is redundant, you can just do:

if (!cacheQuery.canQueryCache)
 +    {
 +        completion(nil, cacheQuery.error);
 +        return;
 +    }

The cacheQuery object will provide an error as appropriate, and canQueryCache should always be NO if there is an error.

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -94,145 +99,295 @@ - (MHVCachedThing *)newThingForRecord:(MHVCachedRecord *)record
...
+         fetchRequest.predicate = predicate;
+         fetchRequest.propertiesToFetch = @[@"xmlString"];
+         fetchRequest.sortDescriptors = @[[NSSortDescriptor sortDescriptorWithKey:@"effectiveDate" ascending:NO]];
+         fetchRequest.fetchLimit = cacheQuery.fetchLimit;
+         
+         NSError *error;
+         NSArray<MHVCachedThing *> *fetchedThings = [self.managedObjectContext executeFetchRequest:fetchRequest error:&error];
+         if (error)
+         {
+             completion(nil, [NSError MHVCacheError:@"Could fetch things from cache database"]);
```
namalu: Could fetch or Couldn't fetch?

HealthVault/Classes/Caching/Private/MHVThingCache.m
```diff
@@ -206,50 +330,49 @@ - (void)prepareCacheForRecords:(MHVRecordCollection *)records
     //Dispatch so app can continue starting while cache is setup
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0),^
```
namalu: Now that these methods are asynchronous, the code within the method implementation should be dispatched on the appropriate queue.

namalu (approved):

PR #107, opened 2017-07-06
Tests for cache and cache database

miburf:
Tests for cache and cache database
Include CoreData generated files


Example/Tests/Caching/MHVThingCacheQueryTests.m
```diff
@@ -0,0 +1,258 @@
...
+    
+    beforeEach(^
+               {
+                   database = nil;
+                   thingCache = nil;
+                   returnedCollection = nil;
+                   returnedError = nil;
+               });
+    
+    context(@"when query record has not synced", ^
```
namalu: I think we can still do a better job naming tests. The point of BDD is that the name of the test provides a clear explanation of exactly what is failing, where it's failing and why. The problem with the name of many of these tests is that a developer will need to hunt around the project to figure out what was tested and eventually why it's broken.

For example, if this particular test fails this is what we see: **MHVThingCache_WhenQueryRecordHasNotSynced_ThingCollectionShouldBeNil**. I'm not exactly sure what method I called that caused this error, which thing collection we are talking about, and what a query record is.

Instead maybe the name should be "when cachedResultsForQueries is called and the cache is not ready" and it "should provide a nil result collection" and it "should provide a detailed error" (We should be checking that the error is the one we expect).

**MHVThingCache_WhenCachedResultsForQueriesIsCalledAndTheCacheIsNotReady_ShouldProvideANilResultCollection**

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -702,16 +702,23 @@ - (NSError *)saveContext
     
     __block NSError *error = nil;
     
-    [self.managedObjectContext performBlockAndWait:^
-     {
-         if ([self.managedObjectContext hasChanges])
+    @try
```
namalu: Why would we need this try?

namalu (changes requested):

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
miburf: I accidentally deleted my real DB password when making some tests.  CoreData throws an exception here if the database is corrupted, catching and returning as error so that doesn't crash the app

namalu (approved):

PR #110, opened 2017-07-10
Personal/namalu/add top skip to thing requests

namalu:


Example/Tests/Caching/MHVCacheQueryTests.m
```diff
@@ -490,14 +490,55 @@
         
         it(@"should set the fetchLimit property", ^
            {
-               [[theValue(cacheQuery.fetchLimit) should] equal:theValue(-1)];
+               [[theValue(cacheQuery.fetchLimit) should] equal:theValue(240)];
```
miburf, line 493: I searched in HealthVault, and yep, it's 240 in the configurations.
"maxFullThingResultsPerGroup 240"

HealthVault/Classes/Clients/MHVThingClient.m
```diff
@@ -165,14 +183,35 @@ - (void)getThingsWithQueries:(MHVThingQueryCollection *)queries
...
+                     if (error)
+                     {
+                         completion(nil, error);
+                     }
+                     else
+                     {
+                         if (resultCollection.count > 0)
+                         {
+                             [results addObjectsFromCollection:resultCollection];
+                         }
```
miburf: I think it should match the query ordering, so query[n] == result[n]

It'll get confusing for users if queries are all cloud, or all cache the ordering matches, but not if there's a mix of cache and cloud queries in one query collection.

HealthVault/Classes/Clients/MHVThingClient.m
```diff
@@ -218,20 +258,34 @@ - (void)getThingsWithQueries:(MHVThingQueryCollection *)queries
...
                      {
+                         MHVPendingThing *thing = result.pendingThings[i];
+                         
                          [keys addObject:thing.key];
+                         
+                         if (keys.count == pendingCount)
+                         {
+                             break;
+                         }
                      }
```
miburf, line 287: nit, I think it'd be easier to read if pendingCount is in the for (,,) part:

for (NSInteger i = 0; i < pendingCount; i++)
{
[keys addObject:result.pendingThings[i + queryForResult.offset]];

HealthVault/Classes/Types/Private/MHVThingQueryResultInternal.h
```diff
@@ -0,0 +1,62 @@
+//
+// MHVThingQueryResultInternal.h
+// MHVLib
```
miburf, line 3: We should do a global replace on this

HealthVault/Classes/Types/Private/MHVThingQueryResultInternal.h
```diff
@@ -0,0 +1,62 @@
...
+@interface MHVThingQueryResultInternal : MHVType
+//
+// Collection of things found
+//
+@property (readwrite, nonatomic, strong) MHVThingCollection *things;
+//
+// If there were too many matches (depends on server quotas & buffer sizes), HealthVault will
+// return only the first chunk of matches. It will also return the keys of the 'pending' things
+// You must issue a fresh query to retrieve these pending things. This is easily done using
+// convenient init methods on MHVThingQuery
```
miburf: The init methods got changed, we may want to add something to make it easier.  getting an array of ids for a MHVPendingThingCollection would do it.

miburf (changes requested):

HealthVault/Classes/Clients/MHVThingClient.m
namalu: I'm going to leave it now, because I know that it works.

HealthVault/Classes/Types/Private/MHVThingQueryResultInternal.h
namalu: A global replace on what?

HealthVault/Classes/Types/Private/MHVThingQueryResultInternal.h
namalu: I updated the comment. Because we are not exposing this publicly anymore, and we are not using the initializer, I don't see the benefit of adding another one we won't be using.

HealthVault/Classes/Types/Private/MHVThingQueryResultInternal.h
namalu: I'll do this in a different PR because it will add 755 files to this review.

miburf (approved):

PR #115, opened 2017-07-12
Implemented creating, updating, and deleting pending operations.

namalu:


HealthVault/Classes/Caching/Private/MHVCachedRecord+Cache.m
```diff
@@ -19,20 +19,31 @@
...
         if ([cachedThing.thingId isEqualToString:thingId])
         {
             return cachedThing;
         }
     }
-#endif
+    return nil;
+}
+
+- (MHVPendingThingOperation *)pendingThingOpertaionWithIdentifier:(NSString *)identifier
```
miburf: nit, Operation

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -759,6 +778,231 @@ - (void)updateLastCompletedSyncDate:(NSDate *_Nullable)lastCompletedSyncDate
...
+        //Create query to filter & order Things
+        NSPredicate *predicate = [NSPredicate predicateWithFormat:@"record.recordId == %@", [recordId lowercaseString]];
+        
+        NSFetchRequest *fetchRequest = [NSFetchRequest fetchRequestWithEntityName:@"MHVPendingThingOperation"];
+        fetchRequest.predicate = predicate;
+        fetchRequest.sortDescriptors = @[[NSSortDescriptor sortDescriptorWithKey:@"originalRequestDate" ascending:NO]];
+        
+        NSError *error;
+        NSUInteger fetchCount = [self.managedObjectContext countForFetchRequest:fetchRequest error:&error];
+        if (error)
```
miburf, line 887: Is it more efficient to get the count and only do the query if it's >0 ?  If so, we may want to add that to the other fetches.  If not, it adds a bit that may not be needed.

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -759,6 +778,231 @@ - (void)updateLastCompletedSyncDate:(NSDate *_Nullable)lastCompletedSyncDate
...
+        {
+            NSArray<MHVPendingThingOperation *> *operations = [self.managedObjectContext executeFetchRequest:fetchRequest error:&error];
+            
+            if (error)
+            {
+                completion(nil, [NSError MHVCacheError:@"Could not fetch things from cache database"]);
+                return;
+            }
+            
+            //Convert cached things back into MHVThings
```
miburf: nit, old comment

HealthVault/Classes/Caching/Private/MHVThingCacheProtocol.h
```diff
@@ -84,17 +85,17 @@ NS_ASSUME_NONNULL_BEGIN
  Adds a method call to the cache to be replayed when the connection is online. This method is called if there is a 'PutThings' or 'RemoveThings' method request when there is no connection to the internet.
 
  @param method MHVMethod The method to be re-issued.
- @param completion Envoked once the process of caching the method has completed or an error occured. NSError error A detailed error about the operation failure.
+ @param completion Envoked once the process of caching the method has completed or an error occured. MHVPendingMethod pendingMethod The pending method that was saved cached. NSError error A detailed error about the operation failure.
```
miburf: nit, "saved cached"

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -759,6 +778,231 @@ - (void)updateLastCompletedSyncDate:(NSDate *_Nullable)lastCompletedSyncDate
...
+            for (int i = 0; i < operations.count; i ++)
+            {
+                MHVPendingThingOperation *operation = operations[i];
+                
+                MHVPendingMethod *pendingMethod = [[MHVPendingMethod alloc] initWithOriginalRequestDate:operation.originalRequestDate
```
miburf, line 909: initWithPendingThingOperation that sets all these properties?

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
```diff
@@ -759,6 +778,231 @@ - (void)updateLastCompletedSyncDate:(NSDate *_Nullable)lastCompletedSyncDate
...
+            }
+            
+        }
+        
+        operation.name = pendingMethod.name;
+        operation.version = pendingMethod.version;
+        operation.originalRequestDate = pendingMethod.originalRequestDate;
+        operation.parameters = pendingMethod.parameters;
+        operation.identifier = pendingMethod.identifier;
+        operation.correlationId = pendingMethod.correlationId.UUIDString;
```
miburf, line 843: these aren't queryable now, but maybe lowercaseString in case they are in the future

miburf (approved):

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
namalu: MHVPendingMethod is a public class - I don't want to create an initializer based on a private class.

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
namalu: I actually think the lowercase string is a bad idea and will be filing a bug to fix it later. Whenever possible, we should be passing NSUUIDs instead of strings. Also, NSUUID.UUIDstring returns a capitalized string - If anything we should be converting to capitalized strings - But in reality we should keep the NSUUID object around until it is stored in the database.

HealthVault/Classes/Caching/Private/MHVThingCacheDatabase.m
namalu: It is quite a bit more efficient to just fetch the count if the result is 0. But we would need to do some extended testing to see if it makes sense to do it everywhere. I think most times there will not be any 'pending methods' so it makes sense here.

PR #117, opened 2017-07-13
Replaced MHVXyzCollection with NSArray<Xyz*>

miburf:
Replaced MHVXyzCollection with NSArray<Xyz*> for best compatibility with Swift


namalu (changes requested):
We shouldn't be doing this. We need to fix the root cause of the issue. We can't ask consumers of the public SDK to use the toArray on every collection. Collections need to work in both Objective-C and Swift.

miburf:
NSArrays are specially bridged to Swift arrays.  There’s no enumeration or other method to implement to get it to work easily.  We're already doing the NSFastEnumeration.

1. We can change MHVCollection to be a subclass of NSMutableArray (which Apple doesn’t recommend). That causes the object type to be lost in a lot of cases in Swift, so doesn't work either. 
Adding the type to a few methods for every collection (so MHVStringCollection adds - (NSString *)objectAtIndex:(NSUInteger)index; and others) works for some methods from Swift, but not all. 
I still haven’t found any way to get the types right so collection.forEach(…) gets the object type if it's an NSArray subclass.

2. Convert every collection to a typed array, MHVThingCollection —> NSArray<MHVThing*>  
That would bridge to swift the best and be future-proof for anything that changes between objective-C and Swift.  That will take some work to move all the current MHVCollection methods to categories, and do the correct serialization.

3. Leave as in this code review, requiring .toArray() from swift.


#2 is a lot of work, but seems like the best solution so that HealthVault is the most useable from Swift.

Example/Tests/Caching/MHVThingCacheSyncTests.m
```diff
@@ -36,11 +36,11 @@
     __block MHVMockDatabase *database =  nil;
     
     MHVPersonInfo *testPerson = [MHVPersonInfo new];
-    testPerson.records = [MHVRecordCollection new];
+    testPerson.records = @[];
```
namalu: It looks like you can just delete this line - the testPerson.records gets set 4 lines later.

HealthVault/Classes/Headers/MHVCommon.h
```diff
@@ -22,3 +22,4 @@
 #import "MHVDictionaryExtensions.h"
 #import "MHVValidator.h"
 #import "MHVDateExtensions.h"
+#import "NSArray+MHVThing.h"
```
namalu: We need to eventually get rid of this file, so we should not add anymore imports. I've seen this imported in a quite few places and it's not a good practice. Ideally, classes should just import what they need.

HealthVault/Classes/Models/Categories/NSArray+Utils.h
```diff
@@ -41,6 +41,8 @@
 
 - (NSArray *)arrayByRemovingObject:(id)object;
 
+- (NSArray *)arrayByRemovingObjectAtIndex:(NSUInteger)index;
```
namalu: Let's not add this to the extensions - It is used in one place. The practice of making a mutable array should be done where needed.

HealthVault/Classes/ThingTypes/MHVCollection.h
```diff
@@ -25,26 +25,26 @@ typedef BOOL (^MHVFilter) (id value);
 // Collections allow objects of a particular type only
 // They enforce the type
 //
-@interface MHVCollection<__covariant ObjectType> : NSObject<NSFastEnumeration>
+@interface MHVCollectionInternal<__covariant ObjectType> : NSMutableArray
```
namalu: This file should be deleted at the end of this exercise. This was the root of all of our issues, and there should be no reason to keep this around.

HealthVault/Classes/Types/MHVThing.m
```diff
@@ -393,206 +393,206 @@ + (MHVThing *)newFromXmlString:(NSString *)xml
...
-    [writer writeElementArray:c_element_thing elements:self.toArray];
-}
-
-- (void)deserialize:(XReader *)reader
-{
-    _inner = [reader readElementArray:c_element_thing asClass:[MHVThing class]];
-}
-
-@end
+//static NSString *const c_element_thing = @"thing";
```
namalu: It looks like all of this code can be deleted?

HealthVault/Classes/Types/MHVThingFilter.h
```diff
@@ -96,13 +96,14 @@
  Only Things that are of the given Type Id(s) contained in the collection will be returned.
  @note The Type Ids in this collection are logically ORed.
  */
-@property (readonly, nonatomic, strong)  MHVStringCollection *typeIDs;
+@property (readwrite, nonatomic, strong)  NSArray<NSString *> *typeIDs;
```
namalu: I'm a little concerned that this changed to readwrite when readonly was fine for this collection before.

HealthVault/Classes/Types/MHVThingFilter.h
```diff
@@ -96,13 +96,14 @@
...
 
 - (instancetype)initWithTypeID:(NSString *)typeID;
 - (instancetype)initWithTypeClass:(Class)typeClass;
 
 @end
 
-@interface MHVThingFilterCollection : MHVCollection<MHVThingFilter *>
+@interface MHVThingFilterCollectionInternal : MHVCollectionInternal<MHVThingFilter *>
```
namalu: This is a public class - We can't append something with 'internal' in a public class. 'Internal' objects need to actually be internal and will should be added to a private directory so SDK consumers will have no access to it.

HealthVault/Classes/Types/MHVThingKey.h
```diff
@@ -40,7 +40,7 @@
 
 @end
 
-@interface MHVThingKeyCollection : MHVCollection<MHVThingKey *> <XSerializable>
+@interface MHVThingKeyCollectionInternal : MHVCollectionInternal<MHVThingKey *> <XSerializable>
```
namalu: Here too. Internal cannot be in a public class.

HealthVault/Classes/Types/Private/MHVThingQueryResultInternal.h
```diff
@@ -54,7 +54,7 @@
 
 @end
 
-@interface MHVThingQueryResultCollectionInternal : MHVCollection<MHVThingQueryResultInternal *>
+@interface MHVThingQueryResultCollectionInternal : MHVCollectionInternal<MHVThingQueryResultInternal *>
```
namalu: This will need to be refactored too. Though this is an internal class we should delete MHVCollectionInternal.

HealthVault/Classes/Utilities/MHVValidator.h
```diff
@@ -158,5 +158,6 @@ while(NO)
                                                 return hr; \
                                             } \
 
-MHVClientResult* MHVValidateCollection(MHVCollection *collection, MHVClientResultCode error);
+MHVClientResult* MHVValidateCollection(MHVCollectionInternal *collection, MHVClientResultCode error);
```
namalu: MHVValidateCollection will need to be removed once MHVCollectionInternal is deleted.

Example/healthvault-ios-sdk/MHVFeaturesConfiguration.m
```diff
@@ -24,19 +24,18 @@ + (MHVConfiguration *)configuration
 {
     MHVConfiguration *config = [MHVConfiguration new];
 
-    config.masterApplicationId = [[NSUUID alloc] initWithUUIDString:@"7c81b435-f091-4444-9534-59eb53da66c9"];
-    config.defaultHealthVaultUrl = [[NSURL alloc] initWithString:@"https://platform.healthvault-ppe.com/platform"];
-    config.defaultShellUrl = [[NSURL alloc] initWithString:@"https://account.healthvault-ppe.com"];
-    config.restHealthVaultUrl = [[NSURL alloc] initWithString:@"https://data.ppe.microsofthealth.net"];
-    config.restVersion = @"1.0-preview2";
+//    config.masterApplicationId = [[NSUUID alloc] initWithUUIDString:@"7c81b435-f091-4444-9534-59eb53da66c9"];
```
namalu: Should this be reverted before merging?

Example/healthvault-ios-sdk/ViewControllers/MHVTypeViewController.m
```diff
@@ -349,7 +350,7 @@ - (void)removeCurrentThing
         
                            @synchronized (self.lockObject)
                            {
-                               [self.things removeObject:selectedThing];
+                               self.things = [self.things arrayByRemovingObject:selectedThing];
```
namalu: We shouldn't be using the SDK extensions in the sample app. I will be making most of the extensions private - also, this method should probably go away.

HealthVault/Classes/Extensions/MHVArrayExtensions.h
```diff
@@ -23,6 +23,12 @@
 
 + (BOOL)isNilOrEmpty:(NSArray *)array;
```
namalu: I'm going to recommend we delete all methods in this file except for isNilOrEmpty. We need to clean up our extensions to include things that really make sense.

1. arrayByAddingObjectsFromArray:atStartingIndex: is used in the sample only, and should not be exposed publicly. 
2. arrayByRemovingObject: should just be done with a mutable array where needed.
3. hasElementsOfType: is not used anywhere.
4. addFromEnumerator: is not used anywhere.

HealthVault/Classes/Models/Categories/NSArray+Utils.h
```diff
@@ -41,6 +41,8 @@
 
 - (NSArray *)arrayByRemovingObject:(id)object;
```
namalu: This whole file needs to be cleaned up. Most of the methods are not used anywhere - They should be deleted. Also let's get rid of the convenience methods like arrayByRemovingObject: We should just do this with a mutable array where needed.

namalu (changes requested):

HealthVault/Classes/Models/Categories/NSArray+Utils.h
miburf: Removed all the unused methods & the one only used once

HealthVault/Classes/Types/MHVThingFilter.h
miburf: Reverted the change.  Old collections were weird too, readonly mutable arrays; added an initWithTypeIDs since you can't change the arrays now.

Example/healthvault-ios-sdk/MHVFeaturesConfiguration.m
miburf: Yes, reverted

HealthVault/Classes/Utilities/MHVValidator.h
miburf: Removed all the MHVCollection references

HealthVault/Classes/Extensions/MHVArrayExtensions.h
miburf: Moved isNileOrEmpty to NSArray+Utils and deleted the file ArrayExtensions file

HealthVault/Classes/Types/MHVThingQuery.m
```diff
@@ -189,20 +202,20 @@ - (void)serialize:(XWriter *)writer
...
     {
-        [writer writeElementArray:c_element_id elements:self.thingIDs.toArray];
+        [writer writeElementArray:c_element_id elements:self.thingIDs];
     }
-    else if (![MHVCollection isNilOrEmpty:self.keys])
+    else if (![NSArray isNilOrEmpty:self.keys])
     {
-        [writer writeElementArray:c_element_key elements:self.keys.toArray];
+        [writer writeElementArray:c_element_key thingName:c_element_thing_id elements:self.keys];
     }
```
miburf: @namalu There's a bug here, it's not serializing the thing keys right.  You can reproduce by turning off caching, and then scrolling down on a page so it reloads the next batch.  Flying out tomorrow morning, and ran out of time to get it fixed. :(  Not serializing the right thing, so the query fails on HV.

namalu (approved):

PR #121, opened 2017-07-14
Add task tracking objects and apis

mikenevi:


HealthVault/Classes/Models/Generated/Categories/MHVTaskTrackingApi.h
```diff
@@ -0,0 +1,49 @@
...
+
+@interface MHVRemoteMonitoringClient (MHVTaskTrackingApi)
+
+/// Post a task tracking
+/// 
+///
+/// @param taskTrackingOccurrence The plan task tracking occurrence.
+/// 
+- (void)taskTrackingPostWithTaskTrackingOccurrence:(MHVTaskTrackingOccurrence* )taskTrackingOccurrence
+    completion:(void(^_Nonnull)(NSObject* _Nullable output, NSError* _Nullable error))completion;
```
miburf, line 43: May be coming, but should a MHV Type, not NSObject for the return

HealthVault/Classes/Models/Generated/MHVTaskTrackingOccurrence.m
```diff
@@ -0,0 +1,82 @@
...
+
+
+/**
+ * Maps json key to property name.
+ * This method is used by `JSONModel`.
+
++ (JSONKeyMapper *)keyMapper {
+  return [[JSONKeyMapper alloc] initWithModelToJSONDictionary:@{ @"taskId": @"taskId", @"trackingDateTime": @"trackingDateTime" }];
+}
+ */
```
miburf, line 54: if JSONModel not used, could remove so it doesn't generate the commented out code

HealthVault/Classes/Models/MHVDateTimeBase.m
```diff
@@ -21,6 +21,17 @@
 
 @implementation MHVDateTimeBase
 
+- (instancetype)init
```
miburf, line 24: I was going to add an initWithDate: since Kayla needed it, if you want to add that too, that'd be nice :)

miburf (approved):

HealthVault/Classes/Models/Generated/Categories/MHVTaskTrackingApi.h
mikenevi: Yeah, the swagger doc isn't quite right for this API call. I'm updating that and will need to regen these once it's fixed & deployed.

HealthVault/Classes/Models/Generated/MHVTaskTrackingOccurrence.m
mikenevi: Yeah, it should be removed. I'll do that later since it will affect all models.

mikenevi: I opened 58529 to do that later (will also fix the NSObject type above).

HealthVault/Classes/Models/MHVDateTimeBase.m
mikenevi: Sure, I'll add it.

PR #124, opened 2017-07-24
Personal/namalu/hide private interfaces

namalu:


HealthVault/Classes/Connection/Private/MHVConnection.m
```diff
@@ -182,8 +182,13 @@ - (void)authenticateWithViewController:(UIViewController *_Nullable)viewControll
 {
     if (!_thingClient)
     {
+#ifdef THING_CACHE
```
miburf: Should do a search and update all to #if THING_CACHE

miburf (approved):

PR #131, opened 2017-08-14
add timeZone param for timeline, remove maxSize params

mikenevi:


Example/healthvault-ios-sdk/ViewControllers/Timeline/MHVTimelineSnapshotViewController.m
```diff
@@ -68,9 +68,10 @@ - (void)loadTimeline
     [self.statusLabel showBusy];
     
     MHVLocalDate* startDate = [[MHVLocalDate alloc] initWithObject:@"2017-05-29" objectParameters:nil];
```
miburf: Maybe update to be the past 60 days or something?  Running the app a couple years from now could be a lot of data

HealthVault/Classes/Models/Generated/Categories/MHVTaskTrackingApi.m
```diff
@@ -40,6 +40,109 @@ @implementation MHVRemoteMonitoringClient (MHVTaskTrackingApi)
...
+{
+    // verify the required parameter 'occurrenceId' is set
+    if (occurrenceId == nil)
+    {
+        NSParameterAssert(occurrenceId);
+        if(completion)
+        {
+            NSDictionary * userInfo = @{NSLocalizedDescriptionKey : [NSString stringWithFormat:NSLocalizedString(@"Missing required parameter '%@'", nil),@"occurrenceId"] };
+            NSError* error = [NSError errorWithDomain:kMHVTaskTrackingApiErrorDomain code:kMHVTaskTrackingApiMissingParamErrorCode userInfo:userInfo];
+            completion(nil, error);
```
miburf, line 60: should return.  This'll then do the request, so would call the completion twice

miburf (approved):
A few things to fix

HealthVault/Classes/Models/Generated/Categories/MHVTaskTrackingApi.m
mikenevi: Don't know how we didn't notice this before 👍

PR #138, opened 2017-08-31
Related Thing helpers, and fix code generation for associatedPlanId 

miburf:
Related Thing helpers, and fix code generation for associatedPlanId

### DIFF
--- a/HealthVault/Tools/codegen/codegen.sh
+++ b/HealthVault/Tools/codegen/codegen.sh
@@ -76,3 +76,9 @@ s/"MHVTrackingValidation.h"/"MHVTrackingValidation.h"\
 #import "MHVPlanStatusEnum.h"/g
 ' output/SwaggerClient/Api/MHVActionPlanTasksApi.m > output/SwaggerClient/Api/MHVActionPlanTasksApi.m2
 mv output/SwaggerClient/Api/MHVActionPlanTasksApi.m2 output/SwaggerClient/Api/MHVActionPlanTasksApi.m
+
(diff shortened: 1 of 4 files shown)


HealthVault/Tools/codegen/codegen.sh
```diff
@@ -76,3 +76,9 @@ s/"MHVTrackingValidation.h"/"MHVTrackingValidation.h"\
 #import "MHVPlanStatusEnum.h"/g
 ' output/SwaggerClient/Api/MHVActionPlanTasksApi.m > output/SwaggerClient/Api/MHVActionPlanTasksApi.m2
 mv output/SwaggerClient/Api/MHVActionPlanTasksApi.m2 output/SwaggerClient/Api/MHVActionPlanTasksApi.m
+
+# Other nullability fixes
+sed -E '
+s/strong,nonatomic) NSString\* associatedPlanId/strong,nonatomic,nullable) NSString* associatedPlanId/g
+' output/SwaggerClient/Model/MHVActionPlanTask.h > output/SwaggerClient/Model/MHVActionPlanTask.h2
```
miburf, line 83: I filed a bug to fix to fix the DTO so Swagger generates it with the nullable attribute, this can be removed when that is fixed

namalu: We should just get the bug fixed and not add this one off code here. You can raise the priority of the bug if this is blocking you in some way, but this is not a good idea.

namalu (changes requested):

